PR #3, opened 2023-04-26
feat: add telemetry infrastructure

nathanielc:
With these change metrics and traces are now collected.

Metrics can be viewed using Prometheus and traces via Jaeger.


k8s/opentelemetry/opentelemetry.yaml
```diff
@@ -0,0 +1,127 @@
...
+            - --config=/config/otel-config.yaml
+          ports:
+            - containerPort: 4317
+              name: otlp-receiver
+            - containerPort: 9090
+              name: prom-metrics
+            - containerPort: 8888
+              name: self-metrics
+          resources:
+            limits:
```
AaronGoldman, line 35: How did we pick these numbers?

AaronGoldman (approved):
Looks like a lot of moving parts but I don't see a way to have fewer of them.

k8s/ceramic/manifests/bootstrap.yaml
```diff
@@ -16,15 +15,15 @@ spec:
           - "/usr/bin/runner"
           - "bootstrap"
         env:
+          - name: RUNNER_OTLP_ENDPOINT
+            value: 'http://otel:4317'
           - name: RUST_LOG
-            value: 'info,runner=debug'
+            value: 'info,runner=trace'
```
dbcfd, line 21: Any reason why runner is at trace?

zachferland (approved):

k8s/ceramic/manifests/bootstrap.yaml
nathanielc: Runner is the code we write so figured it was nice to have full verbosity there. Obviously its easy to change.

PR #4, opened 2023-04-28
feat: triv operator with control loop

zachferland:
Trivial op as described in linear, mostly familiarizing myself with k8s/kube-rs 

TODO 
- seems like i cant get status as a subresource for the crd, api returns 404 from im assuming that resource endpoint, err commented in PR 
- add dockerfile, & run in cluster

### DIFF
--- a/k8s/ceramic/manifests/network.yaml
+++ b/k8s/ceramic/manifests/network.yaml
@@ -1,0 +1,33 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: networks.simulator.io
+spec:
+  group: simulator.io
+  versions:
+    - name: v1
+      subresources:
+        status: {}
+      served: true
+      storage: true
+      schema:
+        openAPIV3Schema:
+          type: object
+          properties:
+            spec:
+              type: object
+              properties:
+                enabled: 
+                    type: boolean
+            status:
+              type: object
+              properties:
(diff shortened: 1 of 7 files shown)


k8s/ceramic/manifests/network.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: apiextensions.k8s.io/v1
```
nathanielc, line 1: Is this file generated using the `kube-rs` tooling?

nathanielc (commented):
Generally looks good, its got the pieces I expected to see. Its not clear what state this is in, can you successfully create a `network` resource and `enable` it? Or does that fail at some point?

zachferland:
@nathanielc  flow that works so far, ` kubectl apply -k ./k8s/ceramic` to create network resource, event comes up in event loop, and on any network resources changes receive additional events. 

do you mean anything else by enable? or another way to create it? 

The part i couldnt get working yet was updating the status, wanted to just update the status based on spec changes, that was the error i included in comment, the resource endpoint for the status wasnt available, even tho i included it in the CRD, but may be missing something or some understanding.

k8s/ceramic/manifests/network.yaml
zachferland: I added this file

nathanielc (approved):
LGTM we can address any remaining issues as we work towards a real operator.

PR #5, opened 2023-05-01
feat: add telemetry infrastructure

nathanielc:
With these change metrics and traces are now collected.

Metrics can be viewed using Prometheus and traces via Jaeger.


nathanielc:
This was already merged

PR #6, opened 2023-05-01
feat: add basic simulator jobs

nathanielc:
Add ability to run basic simulations and obverse/analyze their results. See README for usage.


.github/workflows/pr.yml
```diff
@@ -0,0 +1,10 @@
+---
+name: "Conventional PR and Commit Messages"
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, edited]
+
+jobs:
+  semantic:
+    uses: influxdata/validate-semantic-github-messages/.github/workflows/semantic.yml@main
```
oed, line 10: What does this do exactly? What message format are we enforcing? Should we document this in the readme?

README.md
```diff
@@ -62,3 +73,32 @@ To view the metrics and traces port-forward the services:
...
+> This is a manual method in the meantime.
+
+## Analysis
+
+To analyze the results of a simulation first copy the metrics-TIMESTAMP.parquet file from the otel-0 pod.
+First restart otel-0 pod so it writes out the parquet file footer.
+
+    kubectl delete pod otel-0
+    kubectl cp otel-0:data/metrics-TIMESTAMP.parquet ./analyze/metrics.parquet
+    cd analyze
```
oed, line 94: In your demo you had to do an additional step to find the `TIMESTAMP` should that step be described here?

oed (approved):

.github/workflows/pr.yml
nathanielc: This enforces https://www.conventionalcommits.org/en/v1.0.0/ yeah a note in the README is a good idea. We already use this style of commit message everywhere this just adds a CI check. Helps later if we want to automate generating changelogs.

PR #8, opened 2023-05-10
feat: integrate operator into workspace

nathanielc:
Prior to this change, the operator and runner projects were separate cargo projects. Now they are a singular workspace and the yaml is updated to deploy the operator into k8s.


operator/src/network.rs
```diff
@@ -0,0 +1,99 @@
...
+    let client: Client = context.client.clone();
+    let net_api: Api<Network> = Api::all(client);
+    let name = network.name_any();
+
+    let status = json!({
+        "status": NetworkStatus { enabled: true }
+    });
+
+    // TODO fails, assuming status/resource was not created
+    net_api
```
zachferland, line 67: noticed this was properly formatted/linted now, what should i be using locally during development?

k8s/simulation/network.yaml
```diff
@@ -1,7 +1,7 @@
-apiVersion: "simulator.io/v1"
+apiVersion: "keramik.3box.io/v1"
 kind: Network
 metadata:
-  name: network
-  namespace: ceramic
+  name: bob
```
zachferland, line 4: is the name/namespace set elsewhere?

zachferland (approved):
looks good 👍

operator/src/network.rs
nathanielc: You can configure vscode to format using the rust analyzer on save. That is what I typically do (except with vim)

Alternatively you can run `cargo fmt` to format all files in the workspace.

k8s/simulation/network.yaml
nathanielc: Yes, via kustimozation.yaml, this way the namespace is configurable for all related resources in one place.

PR #10, opened 2023-05-12
feat(operator): adds an operator that can create and bootstrap

nathanielc:
This change represent a move from manually applying yaml resource definitions to use the operator to automatically apply resources.

Generally speaking each yaml file has been converted into a rust file that defines the same resources however now with the ability to programmatically customize parts of each resource.

The simulation logic has been temporarily disabled because the operator made a paradigm shift in how the runner is invoked. Previously the runner assumed a deterministic set of nodes and DNS addresses to those nodes. However with the operator things are a bit more dynamic and its also easier to _assemble_ the set of nodes as opposed to assume their existence deterministically. The bootstrap runner command has been updated according to this new pattern. However to keep the PR smaller the simulation command has not been updated.


nathanielc:
> The simulation logic has been temporarily disabled  

Re-enabled in this PR https://github.com/3box/keramik/pull/11

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+# Copy in source code
+COPY . .
+
+# Build application using a docker cache
+# To clear the cache use:
+#   docker builder prune --filter type=exec.cachemount
+RUN --mount=type=cache,target=/home/builder/.cargo,uid=$UID,gid=$GID \
+	--mount=type=cache,target=/home/builder/keramik/target,uid=$UID,gid=$GID \
+    make build && cp ./target/release/keramik-runner ./target/release/keramik-operator ./
```
nathanielc, line 18: Turns out the `cargo chef` tooling was broken, but I found a better way of getting faster Docker builds using a first class caching feature of docker. This works even better than cargo because its fully incremental instead of loosing all cached deps if any dep changes.

runner/src/bootstrap.rs
```diff
@@ -36,47 +40,58 @@ impl Default for Method {
...
-#[tracing::instrument]
-async fn random(n: usize, total: usize) -> Result<()> {
+#[tracing::instrument(skip(peers), fields(peers.len = peers.len()))]
+async fn random(n: usize, peers: &BTreeMap<usize, PeerInfo>) -> Result<()> {
     let mut rng = rand::thread_rng();
-    let addrs = all_peer_addrs(total).await?;
     // Reuse a peers buffer for each loop
-    let mut peers = vec![0; n];
-    for peer in 0..total {
+    let mut other_peers = vec![0; n];
```
zachferland, line 81: also n or  peers.len() - 1? 
```suggestion
    let mut other_peers = vec![0; min(n, peers.len() - 1)];
```

operator/src/network/cas.rs
```diff
@@ -0,0 +1,631 @@
...
+                            ..Default::default()},
+                        ),
+                    ..Default::default()},
+                ),
+            ..Default::default()},
+        ],
+    ),
+..Default::default()}
+}
+pub fn cas_service_sepc() -> ServiceSpec {
```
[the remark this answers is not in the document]
nathanielc: fixed thanks for catching that

PR #11, opened 2023-05-15
fix: update simulate cmd to take assembled peers info

nathanielc:
Prior to this change the simulate command assumed the peer RPC address based on the peer index. Now with this change a JSON encoding of peers and their address must be specified. This enables the simulate command to work well with the operator.

### DIFF
--- a/runner/src/simulate.rs
+++ b/runner/src/simulate.rs
@@ -1,15 +1,18 @@
-use std::sync::{Arc, Mutex};
+use std::{
+    collections::BTreeMap,
+    fs::File,
+    path::PathBuf,
+    sync::{Arc, Mutex},
+};
 
 use anyhow::{bail, Result};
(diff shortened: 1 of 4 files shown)


runner/src/simulate.rs
```diff
@@ -65,6 +69,79 @@ impl Scenario {
     }
 }
 
+#[tracing::instrument]
+pub async fn simulate(opts: Opts) -> Result<()> {
+    let mut metrics = Metrics::init(&opts)?;
+
+    let f = File::open(opts.peers)?;
```
dbcfd, line 76: It doesn't matter a ton here since it's early in the program, but prefer `tokio::fs` to `std::fs` for any file operations. If the fs calls end up blocking for any amount of time, it can cause issues with the tokio runtime.

dbcfd (approved):

PR #13, opened 2023-05-17
fix: add tests for operator

nathanielc:
These tests follow a basic pattern. First the controller is basically responsible for making API calls to the k8s API. So to test it we use a combination of mocks, stubs and expect_test to build a basic test framework. The mock/stubs capture the requests made to the API and return responses. The `expect_test` crate is used to assert the correctness of the requests while making it easy to update the tests when the code itself is changed. Additionally an `ExpectPatch` concept is added that allows for  patching the default stubs instead of reproducing their entire contents for each test.

### DIFF
--- a/operator/src/network/controller.rs
+++ b/operator/src/network/controller.rs
@@ -788,7 +788,7 @@ mod test {
                    "kind": "ConfigMap",
                    "data": {
             -        "peers.json": "[]"
-            +        "peers.json": "[{\"index\":0,\"peer_id\":\"peer_id_0\",\"rpc_addr\":\"http://peer0\",\"p2p_addrs\":[\"/ip4/10.0.0.1/tcp/4001/p2p/peer_id_0\"]},{\"index\":1,\"peer_id\":\"peer_id_1\",\"rpc_addr\":\"http://peer1\",\"p2p_addrs\":[\"/ip4/10.0.0.2/tcp/4001/p2p/peer_id_1\"]}]"
+            +        "peers.json": "[{\"index\":0,\"peerId\":\"peer_id_0\",\"rpcAddr\":\"http://peer0\",\"p2pAddrs\":[\"/ip4/10.0.0.1/tcp/4001/p2p/peer_id_0\"]},{\"index\":1,\"peerId\":\"peer_id_1\",\"rpcAddr\":\"http://peer1\",\"p2pAddrs\":[\"/ip4/10.0.0.2/tcp/4001/p2p/peer_id_1\"]}]"
                    },
                    "metadata": {
                      "labels": {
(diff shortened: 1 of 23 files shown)


nathanielc:
Incidentally rebasing this PR gave a good opportunity to illustrate what changes to the test stubs look like. In main we now use camelCase names for the spec and status fields as is the k8s convention. The [second  commit](https://github.com/3box/keramik/pull/13/commits/d121a9e7ae94bc8aa1af98fbed6b66ca45ab7e7a) of this PR shows the changes to the test fixtures to update the stubs and their patches with the new camelCase field names.

Besides auditing the stub changes look good, all that was needed was to run `env UPDATE_EXPECT=1 cargo test` to produce the changes for that commit.

operator/src/network/controller.rs
```diff
@@ -30,22 +29,35 @@ use tracing::{debug, error, trace};
...
-    peers, utils, BootstrapSpec, CeramicSpec, Network, NetworkStatus,
+    peers,
+    utils::{HttpRpcClient, RpcClient},
+    BootstrapSpec, CeramicSpec, Network, NetworkStatus,
 };
 
 //can add resource apis here
-struct ContextData {
-    client: Client,
+pub struct Context<R> {
```
zachferland, line 38: what is having both HttpRpcClient and  RpcClient for?

nathanielc: The `rpc_client` is a client to the IPFS RPC endpoint. The `RpcClient` is the trait and `HttpRpcClient` implements the trait (using http). Having the `RpcClient` trait means we can mock it out in tests and not have to spin up IPFS directly. However `IpfsRpcClient` is likely a better name.

operator/src/network/controller.rs
```diff
@@ -685,3 +700,141 @@ async fn apply_config_map(
...
+                })),
+            RpcClientMock::peer_info
+                .next_call(matching!(_))
+                .returns(Ok(PeerInfo {
+                    index: 1,
+                    peer_id: "peer_id_1".to_owned(),
+                    rpc_addr: "http://peer1".to_owned(),
+                    p2p_addrs: vec!["/ip4/10.0.0.2/tcp/4001/p2p/peer_id_1".to_owned()],
+                })),
+        ));
```
nathanielc, line 767: As an example here we mock out the RpcClient to return information about two peers in the network. Then the test validates the networks status contains that information.

operator/src/network/controller.rs
zachferland: ok that makes sense, and the mocks make sense

zachferland (approved):
dont understand everything in network/stub.rs but generally like the expect test/patch setup for the operator and lifecycle for writing/adding/updating tests

PR #14, opened 2023-05-23
feat: initial sim crd

zachferland:
### DRAFT
not ready for any final review, but directional feedback is helpful

### Added:
Adds simulation CRD 
Adds configuration for simulation service, manager, and n works 
Add simulation CRD reconcile loop and runs 

Currently:
- applies sim service and manager if matching network available and network "ready" 
- applies n workers to match n peers from network, and applies when sim manager is ready 

TODO
having some trouble running all network resources, so have not fully tested 
some configs to handle, like nonce 
simulation status 
doesnt handle metrics resources or any post processing/cleanup yet


operator/src/simulation/controller.rs
```diff
@@ -0,0 +1,264 @@
...
+
+    let net_ready = net_status.ready_replicas == net_status.replicas;
+
+    if !net_ready {
+        debug!("simulation waiting, network not ready");
+        return  Ok(Action::requeue(Duration::from_secs(10)))
+    }
+
+    let num_peers = net_status.ready_replicas;
+    let ns = "keramik-".to_owned() + sim_name;
```
nathanielc, line 132: If we make the simulation resource namespaced then we can just use that namespace and not construct it ourselves.

operator/src/network/mod.rs
```diff
@@ -38,8 +38,10 @@ pub struct NetworkSpec {
...
 pub struct NetworkStatus {
-    replicas: i32,
-    ready_replicas: i32,
+    /// Number of Ceramic peers
+    pub replicas: i32,
+    ///  Describes how new peers in the network should be bootstrapped.
+    pub ready_replicas: i32,
     namespace: Option<String>,
     // Information about each Ceramic peer
     peers: Vec<PeerInfo>,
```
nathanielc, line 47: Lets make all the fields public.

operator/src/simulation/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+
+/// Primary CRD for creating and managing a Ceramic Simulation.
+#[derive(CustomResource, Serialize, Deserialize, Debug, PartialEq, Clone, JsonSchema)]
+#[kube(
+    group = "keramik.3box.io",
+    version = "v1",
+    kind = "Simulation",
+    plural = "simulations",
+    status = "SimulationStatus",
+    derive = "PartialEq"
```
nathanielc, line 20: I think we should make the simulation resource namespaced, that way we don't have to find the namespace from the network.

nathanielc (commented):
Overall looks great! I like the refactor to utils for the shared bits. The status/config pattern also worked well in this case too. Added a few small comments below.

zachferland:
Able to start all monitoring resources now with the operator (and build/run all network resources on mac)

zachferland:
Replaced by - https://github.com/3box/keramik/pull/21

PR #17, opened 2023-05-31
feat: adds support for custom IPFS images

nathanielc:
Its now possible to configure custom images for Rust or Go implementations of IPFS. See README for usage details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Place the following network definition into the file `small.yaml`.
 ```yaml
 # small.yaml
 ---
-apiVersion: "keramik.3box.io/v1"
+apiVersion: "keramik.3box.io/v1alpha1"
 kind: Network
 metadata:
   name: small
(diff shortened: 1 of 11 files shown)


nathanielc:
I added this feature to run the smoke tests against Keramik with multiple different IPFS images.

operator/src/network/ceramic.rs
```diff
@@ -116,15 +119,117 @@ pub fn service_spec() -> ServiceSpec {
...
+}
+
+pub struct GoIpfsConfig {
+    image: String,
+    image_pull_policy: String,
+}
+impl Default for GoIpfsConfig {
+    fn default() -> Self {
+        Self {
+            image: "ipfs/kubo:v0.19.1@sha256:c4527752a2130f55090be89ade8dde8f8a5328ec72570676b90f66e2cabf827d". to_owned(),
```
zachferland, line 213: I guess this default we would have have to just update manually? would be nice to have defaults get latest used in ceramic

also is there any config output paired with metrics/results? esp for defaults would be nice to know images used for results

README.md
```diff
@@ -119,8 +118,80 @@ images:
...
+```yaml
+apiVersion: "keramik.3box.io/v1alpha1"
+kind: Network
+metadata:
+  name: example-vanilla-kubo
+spec:
+  replicas: 5
+  ceramic:
+    ipfs:
+      kind: rust
```
zachferland, line 144: defaults are nice 

and this looks like all nodes with same version, eventually mixed networks might be useful

zachferland (approved):
lgtm!

PR #18, opened 2023-05-31
feat: always bootstrap when a peer is isolated

nathanielc:
Prior to this change if a peer restarted it would not bootstrap again,
leaving it isolated. With this change we now explicitly check each peer
for their connected status and bootstrap if any peers are isolated.


smrz2001 (approved):
LGTM! 🚢

PR #20, opened 2023-06-02
feat: ceramic http based scenario

nathanielc:
Adds ceramic scenario where each user creates two streams and then
applies edits to those streams repeatedly.

This PR supersedes https://github.com/3box/keramik/pull/9


operator/src/network/ceramic.rs
```diff
@@ -19,8 +19,8 @@ use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
 use crate::network::controller::{
-    managed_labels, selector_labels, CAS_SERVICE_NAME, CERAMIC_APP, CERAMIC_SERVICE_NAME,
-    CERAMIC_SERVICE_RPC_PORT, GANACHE_SERVICE_NAME, INIT_CONFIG_MAP_NAME,
+    managed_labels, selector_labels, CAS_SERVICE_NAME, CERAMIC_APP, CERAMIC_SERVICE_API_PORT,
```
smrz2001, line 22: I guess I still won't be able to use this image as-is until the CAS is configurable.

smrz2001 (approved):
:shipit:

PR #21, opened 2023-06-05
feat: simulation CRD and operator 

zachferland:
Initial PR - https://github.com/3box/keramik/pull/14

### Added:
Adds simulation CRD
Adds configuration/resources for simulation service, goose manager, and n goose workers
Add simulation CRD reconcile loop and runs
Adds resources for all metrics/monitoring resources

### Currently:
applies all metrics and monitoring resources 
applies sim service and manager if matching network available and network "ready"
applies n workers to match n peers from network, and applies when sim manager is ready

Will be changed to allow configuration over networks not in cluster and to not have dependency on a network to start. 

A number of configuration options can/should be added to the simulation resource still to support more use cases. 

TODO:
tests 
update docs


nathanielc (changes requested):
Looks really good, a few comments below.

operator/src/simulation/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+//   Read,
+//   /// Taking a snapshot of metrics from the network
+//   Snapshot,
+//   /// Remove interested datasets from each node
+//   Cleanup,
+//   /// Simulation is complete, it finished at the specified time
+//   Complete(Time),
+//   /// Simulation failed, and the time of failure
+//   Failed(Time)
+// }
```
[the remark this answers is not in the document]
zachferland: can add the completed/failed times, but didnt seem like we need any other status/state and could add as needed

operator/src/simulation/worker.rs
```diff
@@ -0,0 +1,158 @@
...
+    pub nonce: u32,
+}
+
+// Define clear defaults for this config
+impl Default for WorkerConfig {
+    fn default() -> Self {
+        Self {
+            scenario: "ceramic-simple".to_owned(),
+            target_peer: 0,
+            nonce: 1
```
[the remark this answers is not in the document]
zachferland: yeah agree, workers/managers ended up not really having defaults, can remove that

operator/src/simulation/manager.rs
```diff
@@ -0,0 +1,192 @@
...
+}
+
+// Define clear defaults for this config
+impl Default for ManagerConfig {
+    fn default() -> Self {
+        Self {
+            scenario: "ceramic-simple".to_owned(),
+            users: 100,
+            run_time: 10,
+            nonce: random::<u32>(),
```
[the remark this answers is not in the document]
zachferland: they currently all get the same nonce, i put the nonce in the simulation status and gets a default nonce when created that is passed to all workers/manager 

ended up not using any of the defaults

operator/src/simulation/controller.rs
```diff
@@ -0,0 +1,384 @@
...
+        debug!("simulation waiting, network not ready");
+        return  Ok(Action::requeue(Duration::from_secs(10)))
+    }
+
+    let num_peers = net_status.ready_replicas;
+    let ns = "keramik-".to_owned() + sim_name;
+
+    apply_jaeger(cx.clone(), &ns, simulation.clone()).await?;
+    apply_prometheus(cx.clone(), &ns, simulation.clone()).await?;
+    apply_opentelemetry(cx.clone(), &ns, simulation.clone()).await?;
```
zachferland, line 147: i should probably make sure these are all started/ready before starting workers/managers

operator/src/simulation/controller.rs
```diff
@@ -0,0 +1,384 @@
...
+
+    let net_ready = net_status.ready_replicas == net_status.replicas;
+
+    if !net_ready {
+        debug!("simulation waiting, network not ready");
+        return  Ok(Action::requeue(Duration::from_secs(10)))
+    }
+
+    let num_peers = net_status.ready_replicas;
+    let ns = "keramik-".to_owned() + sim_name;
```
zachferland, line 143: in the other pr you mentioned something about the namespace, was that to just add its own namespace to all the simulation resources (instead of using network one) ? also makes sense once it wont be dependent on a network

nathanielc: Yeah, the simulation resource should live within a namespace instead of outside. By placing it inside a namespace we no longer have to construct the name of the namespace since it will simply be the namespace the resource lives in.

Does that make sense?

zachferland: ah the simulation resource itself, yeah so everything but the simulation resources was in a namespace (the network namespace right now), but simulation should be in a namespace , and let a user just define a namespace in the yaml for a simulation resource?

nathanielc: Yep that's what I am thinking.  I believe you just add the `namespaced` key to the `derive` bits for the resource.

nathanielc (approved):
We should merge now as its in good shape and then we can do the follow up work in smaller PRs/Issues.

operator/src/simulation/controller.rs
zachferland: yeah, that makes sense

zachferland:
made minor changes based on build warnings, trying to get CI to pass

original network crd tests failing, cant really tell from trace what is causing it yet

zachferland:
@nathanielc looks like the merge queue doesnt want to merge it for some reason?

PR #23, opened 2023-06-15
feat: remove sim depend on network, add sim to namespace

zachferland:
Removes simulation dependency on network, fetches peers number from config map in namespace, does not wait for any network related resources to start. 

Also small fixes:
- removes simulator yaml 
- adds line fix to crdgen


zachferland:
Added one more fix commit, working now

nathanielc (approved):

PR #31, opened 2023-06-22
fix: various fixes to make simulator work in AWS

nathanielc:
Needed to explicitly set file permissions and use a non local image name for the runner

### DIFF
--- a/operator/src/simulation/manager.rs
+++ b/operator/src/simulation/manager.rs
@@ -51,8 +51,8 @@ pub fn manager_job_spec(config: ManagerConfig) -> JobSpec {
                 subdomain: Some("goose".to_owned()),
                 containers: vec![Container {
                     name: "manager".to_owned(),
-                    image: Some("keramik/runner:dev".to_owned()),
-                    image_pull_policy: Some("IfNotPresent".to_owned()),
+                    image: Some("public.ecr.aws/r5b3e0r5/3box/keramik-runner:latest".to_owned()),
+                    image_pull_policy: Some("Always".to_owned()),
                     command: Some(vec![
                         "/usr/bin/keramik-runner".to_owned(),
                         "simulate".to_owned(),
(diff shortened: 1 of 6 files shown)


operator/src/simulation/manager.rs
```diff
@@ -51,8 +51,8 @@ pub fn manager_job_spec(config: ManagerConfig) -> JobSpec {
                 subdomain: Some("goose".to_owned()),
                 containers: vec![Container {
                     name: "manager".to_owned(),
-                    image: Some("keramik/runner:dev".to_owned()),
-                    image_pull_policy: Some("IfNotPresent".to_owned()),
+                    image: Some("public.ecr.aws/r5b3e0r5/3box/keramik-runner:latest".to_owned()),
+                    image_pull_policy: Some("Always".to_owned()),
```
smrz2001, line 55: Did we want the `Always` here?

operator/src/simulation/worker.rs
```diff
@@ -32,8 +32,8 @@ pub fn worker_job_spec(config: WorkerConfig) -> JobSpec {
             spec: Some(PodSpec {
                 containers: vec![Container {
                     name: "worker".to_owned(),
-                    image: Some("keramik/runner:dev".to_owned()),
-                    image_pull_policy: Some("IfNotPresent".to_owned()),
+                    image: Some("public.ecr.aws/r5b3e0r5/3box/keramik-runner:latest".to_owned()),
+                    image_pull_policy: Some("Always".to_owned()),
```
smrz2001, line 36: And here?

operator/src/simulation/manager.rs
nathanielc: yes, for the `latest` tag we want `Always` or we will not ever see new updates.

operator/src/simulation/worker.rs
nathanielc: Yep, anytime we use `latest` we also want to use `Always`.

smrz2001 (approved):
🚢

zachferland (approved):

PR #32, opened 2023-06-22
chore: adds stub tests to simulation resource

nathanielc:
With this change we now have stub tests for the simulation resource. Common code between the network and simulation tests has been refactored into utils::test.


operator/src/simulation/stub.rs
```diff
@@ -0,0 +1,345 @@
...
+    ///
+    /// You should await the `JoinHandle` (with a timeout) from this function to ensure that the
+    /// stub runs to completion (i.e. all expected calls were responded to),
+    /// using the timeout to catch missing api calls to Kubernetes.
+    pub fn run(mut self, stub: Stub) -> tokio::task::JoinHandle<()> {
+        tokio::spawn(async move {
+            // We need to handle each expected call in sequence
+
+            // First we handle the call to get the peers config map.
+            self.handle_request_response(stub.peers_config_map.0, &stub.peers_config_map.1)
```
zachferland, line 216: really nice to just be able to list/order the expected calls/resources

zachferland (approved):

PR #33, opened 2023-06-22
feat: add support for resource limits

nathanielc:
With this change CAS (cas, cas_ipfs, ganache and postgres) along with Ceramic nodes (both js and ipfs nodes) can have their resource limits configured.

Currently request == limits.


zachferland (approved):

zachferland:
good to merge? will also make running baselines easier

PR #34, opened 2023-06-27
feat: add sentinel bootstrap method

nathanielc:


runner/src/bootstrap.rs
```diff
@@ -94,3 +97,25 @@ async fn random(n: usize, peers: &BTreeMap<usize, PeerInfo>) -> Result<()> {
     }
     Ok(())
 }
+
+#[tracing::instrument(skip(peers), fields(peers.len = peers.len()))]
+async fn sentinel(n: usize, peers: &BTreeMap<usize, PeerInfo>) -> Result<()> {
+    for peer_info in peers.values() {
+        // Connect to each peer to the first n peers
```
smrz2001, line 104: Should this be `Connect each peer to the first n peers`?

runner/src/bootstrap.rs
```diff
@@ -94,3 +97,25 @@ async fn random(n: usize, peers: &BTreeMap<usize, PeerInfo>) -> Result<()> {
...
+        // Do not attempt to connect to more peers than exist.
+        for sentinel in 0..min(n, peers.len() - 1) {
+            debug!(%peer_info.index, %sentinel, "sentinel peer connection");
+            if let Some(sentinel_info) = peers.get(&sentinel) {
```
smrz2001, line 108: Does it matter that we might call `connect_peers` with the same peer for both sides of the connection?

smrz2001 (approved):
Couple of minor questions but otherwise LGTM!

zachferland (approved):

runner/src/bootstrap.rs
nathanielc: It will pollute logs, as the API call will fail. I fixed this in #35 as I had to rewrite bits of it anyway.

PR #35, opened 2023-06-27
feat: add bootstrap for cas-ipfs node

nathanielc:
With this change the keramik-peers config map will contain both ceramic peers as well as generic IPFS peers.

The bootstrap methods have been updated accordingly.

#34 should merge first.

### DIFF
--- a/operator/src/network/controller.rs
+++ b/operator/src/network/controller.rs
@@ -9,6 +9,7 @@ use k8s_openapi::{
     },
     apimachinery::pkg::util::intstr::IntOrString,
 };
+use keramik_common::peer_info::{CeramicPeerInfo, Peer};
 use kube::{
     api::{DeleteParams, Patch, PatchParams},
     client::Client,
@@ -31,7 +32,7 @@ use crate::network::{
     cas::{self, CasSpec},
     ceramic::{self, CeramicConfig},
     peers,
-    utils::{HttpRpcClient, RpcClient},
+    utils::{ceramic_addr, ceramic_peer_ipfs_rpc_addr, HttpRpcClient, IpfsRpcClient},
     BootstrapSpec, CeramicSpec, Network, NetworkStatus,
 };
 
@@ -44,7 +45,7 @@ use crate::utils::{
 fn on_error(
     _network: Arc<Network>,
     _error: &Error,
-    _context: Arc<Context<impl RpcClient>>,
+    _context: Arc<Context<impl IpfsRpcClient>>,
 ) -> Action {
     Action::requeue(Duration::from_secs(5))
 }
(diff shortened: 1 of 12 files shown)


operator/src/network/controller.rs
```diff
@@ -473,27 +476,52 @@ async fn apply_bootstrap_job(
...
+            p2p_addrs: info.p2p_addrs,
+        }));
     }
+    // ready_replicas count should not include other non ceramic peers
     status.ready_replicas = status.peers.len() as i32;
+
+    // Add extra peers, using negative peer indexes
+    {
+        // CAS IPFS peer
+        let ipfs_rpc_addr = format!("http://{CAS_IFPS_SERVICE_NAME}-0.{CAS_IFPS_SERVICE_NAME}.{ns}.svc.cluster.local:{CAS_SERVICE_IPFS_PORT}");
```
zachferland, line 513: it was mentioned before to have the option of passing a remote CAS, looks like that would still be easy to configure here

operator/src/network/controller.rs
```diff
@@ -511,7 +539,7 @@ async fn update_peer_info(
 }
 
 async fn is_secret_missing(
-    cx: Arc<Context<impl RpcClient>>,
+    cx: Arc<Context<impl IpfsRpcClient>>,
```
zachferland, line 542: 👍

runner/src/bootstrap.rs
```diff
@@ -52,30 +54,34 @@ pub async fn bootstrap(opts: Opts) -> Result<()> {
...
-                    error!(
-                        ?peer_info.index,
-                        ?other_info.index,
-                        ?err,
-                        "failed to bootstrap ring peer"
-                    );
-                }
+        for (_, other) in peers
+            .iter()
+            .cycle()
```
zachferland, line 62: this repeats the iter and is similar to taking mod before?

zachferland (approved):

operator/src/network/controller.rs
```diff
@@ -192,7 +194,8 @@ async fn reconcile(
             delete_job(cx.clone(), &ns, BOOTSTRAP_JOB_NAME).await?;
         } else {
             // We have fewer peers, remove the extra peers from the list
-            status.peers.retain(|peer| peer.index < spec.replicas);
+            // Note since all *extra* peer use negative indexes they should always be retained.
```
zachferland, line 197: didnt understand how the negative indexes are used?

runner/src/bootstrap.rs
nathanielc: Yes, this makes the iteration start back at the beginning when it gets to the end. But now we don't have to do index math like we were before, and since indexes can be negative its a good change.

operator/src/network/controller.rs
nathanielc: Negative indexes are not special. We need an index just to organize and provide an order to all peers in the network (at least from Keramik's perspective). We have two kinds of peers in the network, Ceramic peers and vanilla IPFS peers. For simplicity the implementation assigns indexes starting at 0 and increasing to Ceramic peers and assigns vanilla IPFS peers starting at -1 and decreasing. 

Again this is just an implementation detail but figured I'd call it out in the comments that are comparing indexes to peer counts.

PR #38, opened 2023-06-30
feat: add ability to config ceramic image

nathanielc:
Its now possible to configure the image for the js-ceramic container.


smrz2001 (approved):
:shipit:

PR #39, opened 2023-06-30
fix: stop adding cas-ipfs status with each reconcile

nathanielc:
Prior to this change the cas-ipfs peer info would be added to the network status with each call to reconcile. Additionally the simulation logic did not correctly ignore non Ceramic peers.

### DIFF
--- a/operator/src/simulation/controller.rs
+++ b/operator/src/simulation/controller.rs
@@ -236,7 +236,13 @@ async fn get_num_peers(
     let map = config_maps.get(PEERS_CONFIG_MAP_NAME).await?;
     let data = map.data.unwrap();
     let value = data.get(PEERS_MAP_KEY).unwrap();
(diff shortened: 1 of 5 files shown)


operator/src/simulation/controller.rs
```diff
@@ -236,7 +236,13 @@ async fn get_num_peers(
     let map = config_maps.get(PEERS_CONFIG_MAP_NAME).await?;
     let data = map.data.unwrap();
     let value = data.get(PEERS_MAP_KEY).unwrap();
-    let peers: Vec<Peer> = serde_json::from_str(value).unwrap();
+    let peers: Vec<Peer> = serde_json::from_str::<Vec<Peer>>(value)
```
zachferland, line 239: thanks! ran into this yesterday

zachferland (approved):

PR #44, opened 2023-07-17
feat: add datadog telemetry support

nathanielc:
With this change we can export telemetry data to DataDog.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,3 @@ Book source can then be found in [keramik](./keramik) and can be viewed locally 
 Contributions are welcome! Opening an issue to disucss your idea is a good first step.
 When you are ready please use [convential commit](https://www.conventionalcommits.org/en/v1.0.0/)  messages in your commits and PR titles.
 
-
-
-
-
-
--- a/k8s/networks/small.yaml
+++ b/k8s/networks/small.yaml
@@ -1,6 +1,27 @@
+---
 apiVersion: "keramik.3box.io/v1alpha1"
 kind: Network
 metadata:
   name: small
 spec:
-  replicas: 2
+  replicas: 3
+  datadog:
+    enabled: true
+    version: "testing"
+    profilingEnabled: true
+  ceramic:
+    resourceLimits:
+      storage: 3Gi
+    image: ceramicnetwork/js-ceramic:latest
+    imagePullPolicy: IfNotPresent
+    ipfs:
+      kind: go
+      image: ceramicnetwork/go-ipfs-daemon:latest
+      imagePullPolicy: Always
+
+  # Use custom runner image for bootstrapping
+  bootstrap:
+    image: keramik/runner:dev
+    imagePullPolicy: IfNotPresent
+  cas:
+    image: ceramicnetwork/ceramic-anchor-service:c56be4a3c0bbda012f4b08347e192caf5c405936
--- a/keramik/src/datadog.md
+++ b/keramik/src/datadog.md
@@ -1,0 +1,103 @@
+# Datadog
+
+Keramik can also be configured to send metrics and telemetry data to datadog.
+
+We will be following these steps.
+
+0. Setup Keramik network
+1. Install the operator
+2. Configure the operator
+3. Configure secrets for the operator
+4. Configure datadog within a Keramik network.
+
+The datadog operator needs to be installed into an existing Keramik network.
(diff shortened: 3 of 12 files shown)


README.md
```diff
@@ -340,6 +340,104 @@ To view the metrics and traces port-forward the services:
...
+    kubectl apply -f small.yaml
+
+Install the datadog k8s operator into the network.
+This requires installing `helm`, there doesn't seem to be any other way to install the operator without first installing helm.
+However once the datadog operator is installed helm is no longer needed.
+
+    helm repo add datadog https://helm.datadoghq.com
+    helm install -n keramik-small my-datadog-operator datadog/datadog-operator
+
+Configure the datadog operator:
```
dbcfd, line 367: What do we mean by configure? Do we add this file, or does helm add it for us and we need to adjust it?

dbcfd (changes requested):

README.md
```diff
@@ -340,6 +340,104 @@ To view the metrics and traces port-forward the services:
...
+
+```yaml
+#dd.yaml
+---
+kind: DatadogAgent
```
zachferland, line 372: is this something the network crd/operator could add after if we wanted?

zachferland (approved):

README.md
nathanielc: Maybe, because installing the datadog operator requires helm its not quite clear what is being installed into k8s. As such we would have to do more than just convert this yaml into the operator. But if we find it useful we could figure something out.

README.md
nathanielc: This config file adds an instance of the operator to the namespace. The `helm` commands above just install it (i.e. add the CRDs etc) but do not start the operator running.

3benbox:
These changes seem fine, can we get this merged?

nathanielc:
Rebased and updated docs accordingly. Should be good to go now.

keramik/src/datadog.md
```diff
@@ -0,0 +1,103 @@
+# Datadog
+
+Keramik can also be configured to send metrics and telemetry data to datadog.
+
+We will following these steps.
```
dbcfd, line 5: `We will be`

keramik/src/datadog.md
```diff
@@ -0,0 +1,103 @@
...
+
+0. Setup Keramik network
+1. Install the operator
+2. Configure the operator
+3. Configure secrets for the operator
+4. Configure datadog within a Keramik network.
+
+The datadog operator needs to be installed into an existing Keramik network.
+Create Keramik network using `small.yaml` from this README.
```
dbcfd, line 14: In book now, should link to scenario page.

keramik/src/datadog.md
```diff
@@ -0,0 +1,103 @@
...
+
+    kubectl apply -f small.yaml
+
+Install the datadog k8s operator into the network.
+This requires installing `helm`, there doesn't seem to be any other way to install the operator without first installing helm.
```
dbcfd, line 19: Can we link to helm install?

k8s/networks/small.yaml
```diff
@@ -1,6 +1,27 @@
...
+      kind: go
+      image: ceramicnetwork/go-ipfs-daemon:latest
+      imagePullPolicy: Always
+
+  # Use custom runner image for bootstrapping
+  bootstrap:
+    image: keramik/runner:dev
+    imagePullPolicy: IfNotPresent
+  cas:
+    image: ceramicnetwork/ceramic-anchor-service:c56be4a3c0bbda012f4b08347e192caf5c405936
```
dbcfd, line 27: Should this be latest?

dbcfd (changes requested):

k8s/networks/small.yaml
nathanielc: No latest is broken :( fails to start if it doesn't have AWS creds as it tries to list s3 buckets on startup.

3benbox: I've gotten around this by setting `SQS_QUEUE_URL` to "" in the CAS env.

I don't thins this is comparison is correct https://github.com/ceramicnetwork/ceramic-anchor-service/blob/f0fe6f4a947ec4fffdd48699bae38b8c43112aba/src/services/request-service.ts#L41

dbcfd (approved):

PR #45, opened 2023-07-27
feat: add support for passing additional kubo config

smrz2001:


operator/src/network/controller.rs
```diff
@@ -1706,21 +1702,146 @@ mod tests {
...
+            -                "image": "public.ecr.aws/r5b3e0r5/3box/ceramic-one:latest",
+            -                "imagePullPolicy": "Always",
+            +                "image": "ipfs/kubo:v0.19.1@sha256:c4527752a2130f55090be89ade8dde8f8a5328ec72570676b90f66e2cabf827d",
+            +                "imagePullPolicy": "IfNotPresent",
+                             "name": "ipfs",
+                             "ports": [
+                               {
+                                 "containerPort": 4001,
+            -                    "name": "swarm-tcp",
+            +                    "name": "swarm",
```
nathanielc, line 1786: not part of your change but we should use the same name for both Go and Rust flavors of the swarm port. `swarm-tcp` is the better name IMO but happy with either as long as they are consistent.

nathanielc (approved):
LGTM with one tiny fix (see below). Really like how the tests make it clear what changed.

PR #46, opened 2023-07-31
feat: mdbook for keramik usage

dbcfd:


nathanielc (approved):
Awesome! I was able to serve it locally without issue. Once this merges we can add a link to the github pages URL to the README as well.

PR #49, opened 2023-08-01
feat: add ability to configure simulation job images

nathanielc:


3benbox (approved):
Tested successfully locally with kind

PR #51, opened 2023-08-03
feat: Updates to book based on usage

dbcfd:


keramik/src/deploy_images.md
```diff
@@ -40,53 +40,5 @@ docker buildx build --load -t keramik/runner:dev --target runner .
...
-metadata:
-  name: basic
-  namespace: keramik-small
-spec:
-  scenario: ceramic-simple
-  users: 10
-  run_time: 4
-  image: keramik/runner:dev
-  imagePullPolicy: IfNotPresent
-```
```
nathanielc, line 92: This section should not be deleted. Maybe it can live in a different section be this topic is not covered anywhere else. This section explains how to configure either the Network or Simulation resources to use a local build of the runner.

nathanielc (dismissed):

keramik/src/deploy_images.md
dbcfd: Re-added in its own section now with a couple of links to it https://github.com/3box/keramik/pull/51/files#diff-68393f8f40dbae2e7f5a2f43d4cd93a4d9338f305017f5775a327825db9b7e23

keramik/src/analysis.md
```diff
@@ -8,6 +8,7 @@ To analyze the results of a simulation first copy the metrics-TIMESTAMP.parquet
 First restart opentelemetry-0 pod so it writes out the parquet file footer.
 
     kubectl delete pod opentelemetry-0
+    sleep 5 # Give pod time to restart
```
3benbox, line 11: Just a note, you can use the api the know when the pod is ready
```
kubectl wait --for=condition=Ready pod/opentelemetry-0
```

3benbox (approved):
lgtm

keramik/src/analysis.md
dbcfd: Thanks, will add that in the next round of book changes

PR #52, opened 2023-08-03
fix: users check should only be in manager

dbcfd:


nathanielc:
This doesn't look quite right. The workers get the number of users they should run from the manager when they start up. You should see in the manager logs how many users it assigns to each worker.

3benbox:
> This doesn't look quite right. The workers get the number of users they should run from the manager when they start up. You should see in the manager logs how many users it assigns to each worker.

The manager and spec match.

```
$ kubectl get pod simulate-manager-8wg42 -o yaml | yq e '.spec.containers[0].env[] |select(.name == 
"SIMULATE_USERS")'
name: SIMULATE_USERS
value: "333"

$ kubectl get simulations js-ceramic-test-users-sim -o yaml | yq e '.spec'
run_time: 5
scenario: ceramic-simple
users: 333

```

Manager logs look ok

```
$ kubectl logs simulate-manager-8wg42
2023-08-03T20:01:36.806024Z  INFO simulate: goose::manager: manager listening on tcp://0.0.0.0:5115, waiting for 3 workers     opts=Opts { scenario: CeramicSimple, manager: true, target_peer: 0, peers: "/keramik-peers/peers.json", users: 333, run_time: "5m", nonce: 2991874900 }
2023-08-03T20:01:36.806043Z  INFO simulate: goose::manager: each worker to start 111 users     opts=Opts { scenario: CeramicSimple, manager: true, target_peer: 0, peers: "/keramik-peers/peers.json", users: 333, run_time: "5m", nonce: 2991874900 }
```

The worker is getting 100?
```
$ kubectl logs simulate-worker-0-6k7lm
2023-08-03T20:01:39.093170Z  INFO keramik_runner: starting runner args.command=Simulate(Opts { scenario: CeramicSimple, manager: false, target_peer: 0, peers: "/keramik-peers/peers.json", users: 100, run_time: "10m", nonce: 2991874900 }) args.otlp_endpoint="http://otel:4317"
Error: number of users 100 must be a multiple of the number of peers 3, this ensures we can deterministically identifiy each user
```

nathanielc:
> The worker is getting 100?

The worker gets is users from the manager. The `users` argument into the work process is meaningless. (We should probably make it an error if it supplied). So that `100` should be ignored by the worker. But there is a bug.

This check https://github.com/3box/keramik/blob/main/runner/src/simulate.rs#L106 should only be done if its the manager process because only the manager honors the `users` argument.

That if statement should read:
```
if opts.manager && opts.users % peers.len() != 0 {
    //...
}
```

PR #53, opened 2023-08-03
fix: cas wallet address and sqs queue name

3benbox:
The latest CAS images fail with an AWS_DEFAULT_REGION missing error because it's trying to use SQS, it should not be trying to use SQS.

Setting the `SQS_QUEUE_URL` to `""` causes it to skip SQS steps and operate locally.

I'm not sure where the existing `ETH_WALLET_PK` came from, but it isn't working.
I took the first address from the ganache output for the new key, it has been consistent between runs.

### DIFF
--- a/operator/src/network/cas.rs
+++ b/operator/src/network/cas.rs
@@ -133,6 +133,11 @@ pub fn cas_stateful_set_spec(
     let config = config.into();
     let mut env = vec![
         EnvVar {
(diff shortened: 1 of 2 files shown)


smrz2001 (approved):
🥳

3benbox:
Closing this for https://github.com/3box/keramik/pull/54
SQS should be fixed upstream

PR #58, opened 2023-08-15
fix: metric annotations dont scale, use endpoints

3benbox:
This PR changes the otel collector jobs to only concern itself with services in it's own namespace.

The current method of using labels doesn't work when there are multiple containers in 1 pod.

The method in the PR defines the container names, ports and paths for metric collection.

I'm not really happy with this method either and think we should consider having the operator expose a [Prometheus HTTP SD](https://prometheus.io/docs/prometheus/latest/http_sd/) compatible endpoint. 

In this scenario, otel and/or prometheus would be configured with a single target of something like http://keramik-operator.keramik/?namespace=<OTEL's NAMESPACE>


3benbox:
Closing for another PR

PR #59, opened 2023-08-17
feat: query scenario

dbcfd:
Implement a query scenario. To properly test query scenario, implemented a method to report logic errors as well when running goose tests.

### DIFF
--- a/runner/src/scenario/ceramic/mod.rs
+++ b/runner/src/scenario/ceramic/mod.rs
@@ -1,45 +1,46 @@
 mod models;
 pub mod new_streams;
+pub mod query;
+mod util;
 pub mod write_only;
 
+use crate::goose_try;
+use crate::scenario::ceramic::util::{goose_error, setup_model, setup_model_instance};
 use ceramic_http_client::api::StreamsResponseOrError;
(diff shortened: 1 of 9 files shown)


dbcfd:
Tested and working in keramik

runner/src/scenario/ceramic/mod.rs
```diff
@@ -130,28 +130,41 @@ async fn update_small_model(user: &mut GooseUser) -> TransactionResult {
...
+    let mut goose = user
+        .request(
+            GooseRequest::builder()
+                .method(GooseMethod::Post)
+                .set_request_builder(req)
+                .expect_status_code(200)
+                .build(),
+        )
+        .await?;
+    let resp: StreamsResponseOrError = goose.response?.json().await?;
```
smrz2001, line 142: Is the `?` at the end desired if we want the `goose_try` macro to resolve both `Ok` and `Err` results?

IIUC, we'll return early if there's an error and so the macro will only ever see successful results.

dbcfd: This one is expected, since this would be the error if goose can't convert the response to json and the expected type.

smrz2001 (approved):
🥳

PR #69, opened 2023-08-24
feat: add ability to configure multiple ceramics

nathanielc:
With this change it is now possible to create multiple Ceramic specs within the same network. The replicas of the network will be spread out across the Ceramic specs. This enables us to configure test networks with half Rust and half Go IPFS implementations as well as mixed js-ceramic versions. See book for more usage details.


operator/src/network/controller.rs
```diff
@@ -157,6 +157,8 @@ pub const CERAMIC_LOCAL_NETWORK_TYPE: &str = "local";
 
 pub const BOOTSTRAP_JOB_NAME: &str = "bootstrap";
 
+const MAX_CERAMICS: usize = 10;
```
dbcfd, line 160: Why is there a max number of ceramics?

nathanielc: Keramik needs to clean up after itself if existing ceramics are removed. (i.e. network is configured with both a rust and go set, when the go set is removed it needs to be deleted). The max 10 is a simple way to do that, either its defined or its not, the ones that are not defined are deleted. k8s is inherently racy so simply storing in the status which ceramics had been previously defined doesn't provide a clean path to deleting old ones. 

And having 10 different configurations should be plenty for current use cases.

A better long term solution would be to use labels and query all resources with the matching labels and then delete the extra ones, but having a limit was a simple enough for now. However I could be convinced to do the labels logic now if we think that's better.

dbcfd (approved):

operator/src/network/ceramic.rs
```diff
@@ -141,17 +145,6 @@ pub struct CeramicSpec {
     pub ipfs: Option<IpfsSpec>,
     /// Resource limits for ceramic nodes, applies to both requests and limits.
     pub resource_limits: Option<ResourceLimitsSpec>,
-    /// Name of secret containing the private key used for signing anchor requests and generating
-    /// the Admin DID.
-    pub private_key_secret: Option<String>,
```
smrz2001, line 146: There's a bit of a problem with this (independent of this PR and mostly my fault). Private keys/DIDs are not supposed to be shared across Ceramic nodes - each is supposed to use its own to prevent anchoring conflicts in case both nodes write to the same stream.

smrz2001: We'd need to generate fresh keys for each node when the secret is unspecified and find a way to automatically authenticate with CAS Auth. We could also allow specifying a key per node but that could get messy and we'd need to check that keys don't collide, etc.

Feels like cleanest, easiest would be to generate and auth new keys. That would bloat the API key space but I don't think that's too much of an issue.

nathanielc: ok, want to create an issue for that? Let fix it in a separate PR

smrz2001: Yep: https://linear.app/3boxlabs/issue/RES-1206/create-and-authenticate-dids-for-keramik-ceramic-nodes

smrz2001 (approved):
🥳

PR #73, opened 2023-08-31
fix: add missing camelCase attribute

nathanielc:
A few of the new top level fields on the Network spec were using underscores. This fixes then to use camelCase.


smrz2001 (approved):
:shipit:

PR #75, opened 2023-09-01
feat: Add redis to keramik

dbcfd:


operator/src/network/redis.rs
```diff
@@ -0,0 +1,85 @@
...
+                    lbls.append(&mut managed_labels().unwrap());
+                    lbls
+                }),
+                ..Default::default()
+            }),
+            spec: Some(PodSpec {
+                containers: vec![Container {
+                    name: REDIS_APP.to_owned(),
+                    image: Some("redis:latest".to_owned()),
+                    image_pull_policy: Some("Always".to_string()),
```
smrz2001, line 56: Not a big deal but should we change this to IfNotPresent since it's unlikely to change much?

smrz2001 (approved):
🚀

nathanielc (changes requested):
Change looks good, However I have a design question.

Why is the redis service/stateful set added to the network CRD and not the simulation CRD? 

Seems like the use case is to allow worker jobs to communicate global state via redis. Do we need redis to be part of the network in that case? 

Is it because we want Redis to outlive the simulation? If so I'd like to understand that case. 

As we move keramik closer to being used to deploy production networks I want to make sure the performance testing use case remains well isolated from the deployment use case.

dbcfd:
> Change looks good, However I have a design question.
> 
> Why is the redis service/stateful set added to the network CRD and not the simulation CRD?
> 
> Seems like the use case is to allow worker jobs to communicate global state via redis. Do we need redis to be part of the network in that case?
> 
> Is it because we want Redis to outlive the simulation? If so I'd like to understand that case.
> 
> As we move keramik closer to being used to deploy production networks I want to make sure the performance testing use case remains well isolated from the deployment use case.

I went back and forth on this, but it felt more like it belonged at the network layer, given that ceramic/CAS live at the network layer, and this is a service that simulations depend on. I could also see the services actually using redis at some point for caching/sharing state. 

Should it be optional at network layer, or just part of simulation layer?

operator/src/network/stub.rs
```diff
@@ -66,6 +66,8 @@ pub struct Stub {
     pub cas_ipfs_stateful_set: ExpectPatch<ExpectFile>,
     pub ganache_stateful_set: ExpectPatch<ExpectFile>,
     pub cas_postgres_stateful_set: ExpectPatch<ExpectFile>,
+    pub ceramic_configmaps: Vec<ExpectPatch<ExpectFile>>,
+    pub ceramic_service: ExpectPatch<ExpectFile>,
```
nathanielc, line 70: These fields are never used. Should we remove them?

nathanielc (approved):
LGTM with the left over changes to the network stub removed.

PR #86, opened 2023-09-14
feat: add RUST_LOG support

nathanielc:


operator/src/network/ceramic.rs
```diff
@@ -164,6 +164,8 @@ pub struct RustIpfsSpec {
     pub image_pull_policy: Option<String>,
     // Resource limits for ipfs nodes, applies to both requests and limits.
     pub resource_limits: Option<ResourceLimitsSpec>,
+    /// Value of the RUST_LOG env var.
+    pub rust_log: Option<String>,
```
3benbox, line 168: What happens on an invalid string?
Does it use default, crash, not log?

3benbox: invalid meaning somehting like `"invalid-logs-level,ceramic_onehundred=debug,tracing_actix_web=all"`

nathanielc: Defaults to `info` for all logs.

3benbox (approved):
lgtm

PR #88, opened 2023-09-18
chore: update deps and expose IPFS types

nathanielc:
This PR makes it possible to consume the keramik operator as a dependency in Rust code. Specifically we needed to update some deps (namely kube) and expose some types.


smrz2001 (approved):
🐣

PR #89, opened 2023-09-20
refactor: create spec only crate without default features

nathanielc:
With this change when you use keramik-operator as a dependency without its default features you get a crate that only contains the spec types with minimal deps.


smrz2001 (approved):
LGTM! 🥳

PR #90, opened 2023-09-21
chore: breakup CI workflow for faster runs

nathanielc:
Also adds sscache support that has been working well in other rust projects


smrz2001 (approved):
:shipit:

PR #92, opened 2023-09-22
feat: wait for pod ready

nathanielc:
Prior to this change keramik only checked the Ceramic API. Now its checks the pod ready condition.


smrz2001 (approved):
🚀

PR #94, opened 2023-10-06
feat: add network TTL

nathanielc:
Now that we are programmatically creating networks (see ceramic-tests) we want to make sure we are not leaving lots of unused resources around.

With this feature the test driver can create networks with a generous TTL of a few hours (when we expect tests to run in a few minutes). If for some reason the test driver fails to clean up the network itself, eventually the operator will clean it up.

This feature remains optional so manually created networks are not affected.

### DIFF
--- a/operator/src/network/controller.rs
+++ b/operator/src/network/controller.rs
@@ -2,10 +2,13 @@ use std::{cmp::min, collections::BTreeMap, str::from_utf8, sync::Arc, time::Dura
 
 use anyhow::anyhow;
 use futures::stream::StreamExt;
-use k8s_openapi::api::{
-    apps::v1::{StatefulSet, StatefulSetStatus},
-    batch::v1::Job,
-    core::v1::{ConfigMap, Namespace, Pod, Secret, Service, ServiceStatus},
+use k8s_openapi::{
+    api::{
+        apps::v1::{StatefulSet, StatefulSetStatus},
+        batch::v1::Job,
+        core::v1::{ConfigMap, Namespace, Pod, Secret, Service, ServiceStatus},
+    },
+    apimachinery::pkg::apis::meta::v1::Time,
 };
 use keramik_common::peer_info::{CeramicPeerInfo, Peer};
 use kube::{
@@ -23,7 +26,7 @@ use kube::{
     Resource,
 };
 use rand::RngCore;
-use tracing::{debug, error, trace, warn};
+use tracing::{debug, error, info, trace, warn};
 
 use crate::{
     labels::{managed_labels, MANAGED_BY_LABEL_SELECTOR},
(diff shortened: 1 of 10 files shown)


operator/src/network/controller.rs
```diff
@@ -183,12 +186,34 @@ async fn reconcile(
...
     };
 
-    let datadog: DataDogConfig = (&spec.datadog).into();
+    // Check if the network should die, otherwise update ttl_seconds_remaining.
+    let creation_timestamp = network.meta().creation_timestamp.as_ref();
+    status.ttl_seconds_remaining = if let Some(ttl_seconds) = &spec.ttl_seconds {
+        if let Some(created) = creation_timestamp {
+            let now = Utc::now();
+            let elapsed = now - created.0;
+            let ttl_seconds_remaining = elapsed.num_seconds() - ttl_seconds;
```
stbrody, line 195: isn't this arithmetic backwards?  Shouldn't it be tll - elapsed, not elapsed - ttl?

operator/src/network/controller.rs
```diff
@@ -2974,4 +3003,30 @@ mod tests {
...
+        let mock_rpc_client = MockIpfsRpcClientTest::new();
+
+        let (testctx, api_handle) = Context::test(mock_rpc_client);
+        let fakeserver = ApiServerVerifier::new(api_handle);
+        let mut network = Network::test().with_spec(NetworkSpec {
+            // Set expiration as now
+            ttl_seconds: Some(0),
+            ..Default::default()
+        });
+        // Set creation time as now.
```
stbrody, line 3019: can we add a test where ttl and creation time aren't the same so we make sure the arithmetic order/sign is correct?

operator/src/network/spec.rs
```diff
@@ -57,6 +60,9 @@ pub struct NetworkStatus {
     pub namespace: Option<String>,
     /// Information about each Ceramic peer
     pub peers: Vec<Peer>,
+    /// The number of seconds this network has left to live.
+    /// If unset the network lives forever.
+    pub ttl_seconds_remaining: Option<i64>,
```
stbrody, line 65: instead of tracking the number of seconds remaining, can we convert the ttl_seconds into a `expires_at` timestamp first thing at startup, so then we only ever deal in absolute times not relative times?  I generally have found dealing with relative times to be pretty error-prone and easy to mess up, converting to absolute timestamps at the edges as soon as possible tends to help keep things simpler and avoid bugs down the road.

stbrody (changes requested):

operator/src/network/controller.rs
nathanielc: done, added both an expire and not expired test

smrz2001 (approved):
🚢

stbrody (approved):
lgtm

PR #98, opened 2023-10-27
fix: update Dockerfile to use correct upstream parent

nathanielc:
Rust-builder no longer publishes the debian-latest tag and uses the debian flavor for its latest. This updates our Dockerfile to use the correct upstream parent image so we do not get glibc version conflicts.


smrz2001 (approved):
:shipit:

AaronGoldman (approved):

PR #108, opened 2023-11-08
fix: move persistant path for localstack to /var/lib

3benbox:
Built and pushed, seems to be working. There is data in the mapped path even with the pinned image version.


smrz2001 (approved):
:shipit:

PR #118, opened 2023-12-07
fix: fixes bug where nodes were connecting to clay

nathanielc:
The changes to rust-ceramic that use the network to connect to a predefined set of nodes based on the network meant that the CAS ipfs node was connecting to clay and therefore the rest of the nodes would learn about and connect to clay.

This changes is so that CAS uses a shared IPFS spec with Ceramic, thus fixing the bug that the CAS ipfs network was not being configured.

This is a breaking change in that the pubsub_topic is no longer needed and the ipfs_resource_limits is removed from CAS as its not part of `ipfs` spec directly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2640,7 +2640,6 @@ dependencies = [
  "goose",
  "keramik-common",
  "libipld 0.16.0",
- "multibase 0.9.1",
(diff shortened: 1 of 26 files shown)


smrz2001 (approved):
🚀

PR #119, opened 2023-12-11
fix: env var changed for cas metrics

3benbox:
Upstream changed the env here https://github.com/ceramicnetwork/ceramic-anchor-service/pull/1177


smrz2001 (approved):
🚀

PR #120, opened 2023-12-13
feat: Add workflow to build images on demand

3benbox:
The GCP route doesn't work with our Dockerfile mount cacheing (they have their own cacheing) so this seems easier.


smrz2001 (approved):
🚀

PR #126, opened 2024-01-09
feat: Add Recon event sync scenarios

dav1do:
Linear ticket WS1-1409.

This PR includes #127 and pulls those changes into main, as well as some refactoring, and the implementation of two new scenarios called `recon-event-key-sync` and `recon-event-sync`. These scenarios include a step to inspect the goose metrics, as well as metrics from other peers to verify that the scenario passed. It will return a non-zero exit code if fails. This looks different to the manager process than an execution error, however, I'm not sure what needs to be done to prevent the manager from restarting in this case. As soon as it completes with the error, a new job starts but it won't actually start a new simulation (I hope/think this is okay).

The simulation spec includes a new field `success_request_target` that will control the threshold used to verify the scenario passed. Here is my thinking about how it should work from the field docs: `Request target for the scenario to be a success. Scenarios can use this to validate throughput and correctness before returning. The exact definition is left to the scenario (requests per second, total requests, rps/node etc).`

I imported rust-ceramic core to generate Event IDs and had to adjust some multiformat versions. I tried to avoid updating things or making too many changes, but the newest version of multihash refactors things substantially. It isn't backward compatible and hasn't been pulled into ipld-dag-cbor yet, but is in new multiaddr and CID crates, so we use public exports and a shared version when appropriate.

---- 
**New Scenario**

This adds two new runner scenarios for Recon. They both follow a similar flow, however, one sends `{eventId: "key", eventData: "payload"}` while the other just sends event IDs.  The flow is as follows, by default it targets 300 req/second, and can be overridden by the `sucessRequestTarget` scenario spec field. This target is verified as number of new events created on every peer using the `recon_key_insert_count_total` counter.

This scenario is a bit odd in that it mostly talks to the IPFS endpoint for recon, but also needs the ceramic API for model creation. We also constrain which node and how many users take certain actions in a way we historically haven't.

**Setup** (once):

- One user on one worker creates a model. It publishes the model ID to redis for other nodes.
  - Goose doesn't provide APIs to communicate from manager/workers currently, so we use redis. There is a socket that workers connect and report to the manager so we could implement it but redis is easy enough. Currently, workers are assigned a unique ID (int) when connecting to the manager, which we use to ensure that only 1 worker creates a model and creates events.
- One user on every workers will subscribe to updates for this model (after reading it out of redis).

**Key sync** (every time):
- One worker will create a random new event ID and store it. This triggers recon sync. We have all the users on the worker doing this, in order to create as much load as we desire
- The other workers no-op.

```yaml
kind: Simulation
metadata:
  name: recon
  # Must be the same namespace as the network to test
  namespace: keramik-small
spec:
  scenario: recon-event-sync # or recon-event-key-sync if you're using rust-ceramic without recon data
  throttleRequests: 280 # include to define a "steady sync" scenario where the worker creates about this many events/second
  successRequestTarget: 10 # override the default value of 300 to determine how many events must be synced per second to other peers in the network
  users: 8 # increase to generate more events, 1 is probably sufficient for our 300/s goal
  runTime: 10
  ```

### DIFF
--- a/runner/Cargo.toml
+++ b/runner/Cargo.toml
@@ -8,9 +8,9 @@ edition = "2021"
 
 [dependencies]
 anyhow.workspace = true
+ceramic-core.workspace = true
 ceramic-http-client = { git = "https://github.com/3box/ceramic-http-client-rs.git", branch = "main", default-features = false }
 #ceramic-http-client = { path = "../../ceramic-http-client-rs", default-features = false }
-cid = "0.9"
 clap.workspace = true
 goose = { version = "0.16", features = ["gaggle"] }
 keramik-common = { workspace = true, features = ["telemetry", "tokio-console"] }
--- a/runner/Cargo.toml
+++ b/runner/Cargo.toml
@@ -8,6 +8,7 @@ edition = "2021"
 
 [dependencies]
 anyhow.workspace = true
+async-trait.workspace = true
 ceramic-core.workspace = true
 ceramic-http-client = { git = "https://github.com/3box/ceramic-http-client-rs.git", branch = "main", default-features = false }
 #ceramic-http-client = { path = "../../ceramic-http-client-rs", default-features = false }
@@ -27,3 +28,6 @@ tokio.workspace = true
 tracing-log.workspace = true
 tracing.workspace = true
 multibase.workspace = true
+
+[dev-dependencies]
+tracing-test = "0.2"
--- a/runner/Cargo.toml
+++ b/runner/Cargo.toml
@@ -27,7 +27,8 @@ serde_json.workspace = true
 tokio.workspace = true
 tracing-log.workspace = true
 tracing.workspace = true
+tracing-subscriber.workspace = true
 multibase.workspace = true
 
 [dev-dependencies]
-tracing-test = "0.2"
+test-log = "0.2"
--- a/runner/src/scenario/ceramic/event_id_sync.rs
+++ b/runner/src/scenario/ceramic/event_id_sync.rs
@@ -10,7 +10,7 @@ use goose::prelude::*;
 use libipld::cid;
 use multihash::{Code, MultihashDigest};
 use reqwest::Url;
-use std::sync::atomic::AtomicBool;
+use std::sync::atomic::{AtomicBool, AtomicU64};
 use std::{sync::Arc, time::Duration};
 use tracing::{info, instrument};
 
@@ -24,6 +24,7 @@ pub(crate) const CREATE_EVENT_TX_NAME: &str = "create_new_event";
 pub(crate) const CREATE_EVENT_REQ_NAME: &str = "POST create_new_event";
 
 static FIRST_USER: AtomicBool = AtomicBool::new(true);
+static NEW_EVENT_CNT: AtomicU64 = AtomicU64::new(0);
 
 fn should_request_events() -> bool {
     goose::get_worker_id() == 1
@@ -109,11 +110,15 @@ async fn setup(
     user.set_session_data(user_data);
     user.base_url = Some(ipfs_peer_addr); // Recon is only available on IPFS address right now
 
-    let _subscribed_to_models = user
+    let request_builder = user
         .get_request_builder(&GooseMethod::Get, &path)?
-        .timeout(std::time::Duration::from_secs(5))
-        .send()
-        .await?;
+        .timeout(Duration::from_secs(5));
+    let req = GooseRequest::builder()
+        .set_request_builder(request_builder)
+        .expect_status_code(200)
+        .build();
+
+    let _goose = user.request(req).await?;
 
     Ok(())
 }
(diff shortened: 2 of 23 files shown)

### CODE REVIEW

File: runner/src/simulate.rs [hunk 1]
```diff
@@ -95,57 +116,244 @@ impl Scenario {
...
+        }
+
+        match self.scenario {
+            Scenario::IpfsRpc
+            | Scenario::CeramicSimple
+            | Scenario::CeramicWriteOnly
+            | Scenario::CeramicNewStreams
+            | Scenario::CeramicQuery
+            | Scenario::CeramicModelReuse => CommandResult::Success,
+            Scenario::EventIdSync => {
```
dav1do, line 250: This currently always fails with 0 events created on other nodes. I need to figure out if I'm subscribing incorrectly or something else is going wrong. I think I can investigate that while this is reviewed and either fix here or as a follow up. The existing scenarios work as before, and this scenario is being iterated on so having it 100% correct right away isn't required imo.

dav1do: Found at least part of the problem.. working on testing the fix

dav1do: Okay, fixed some things but it still doesn't quite work.. the metric is **always** one on the other node, and it says they are in sync. I ran scenarios creating events slowly, quickly, and it always happened.

runner/src/simulate.rs
```diff
@@ -95,58 +117,327 @@ impl Scenario {
...
+                    self.before_metrics = Some(res);
+                    Ok(())
+                }
+            }
+        }
+    }
+
+    /// For now, most scenarios are successful if they complete without error and only EventIdSync has a criteria.
+    /// Not a result to ensure we always proceed with cleanup, even if we fail to validate the scenario.
+    async fn validate_scenario_success(&self, metrics: &GooseMetrics) -> CommandResult {
```
dav1do, line 291: I started to refactor things a bit to make this more testable (e.g. inject a metrics collector trait) and write a couple tests. I'm planning to push that through in a follow up PR so we're not blocked by having the scenario.

runner/src/simulate.rs
```diff
@@ -95,58 +118,391 @@ impl Scenario {
...
-    let mut metrics = Metrics::init(&opts)?;
+type MetricsCollector = Box<dyn MetricsCollection + Send + Sync>;
 
-    let peers: Vec<Peer> = parse_peers_info(opts.peers)
-        .await?
-        .into_iter()
-        .filter(|peer| matches!(peer, Peer::Ceramic(_)))
-        .collect();
+#[async_trait::async_trait]
+trait MetricsCollection: std::fmt::Debug {
```
dav1do, line 138: I probably should have attached this to the existing metrics struct.. would like to revisit to hopefully simplify/clean up the validation a bit, but might punt for now...

runner/src/scenario/ceramic/event_id_sync.rs
```diff
@@ -0,0 +1,182 @@
...
+}
+
+const SORT_KEY: &str = "model";
+// hard code test controller in case we want to find/prune later
+const TEST_CONTROLLER: &str = "did:key:z6MkoFUppcKEVYTS8oVidrja94UoJTatNhnhxJRKF7NYPScS";
+
+fn random_event_id(sort_value: &str) -> ceramic_core::EventId {
+    let cid = random_cid();
+    EventId::new(
+        &ceramic_core::Network::Local(42),
```
nathanielc, line 174: if this doesn't match the value of the configured nodes then it will not synchronize. I believe we use 0.

runner/src/scenario/ceramic/event_id_sync.rs
```diff
@@ -0,0 +1,182 @@
...
+    let _goose = user.request(req).await?;
+
+    Ok(())
+}
+
+/// Generate a random event that the nodes are interested in. Only one node should create but all
+/// users do it so that we can generate a lot of events.
+async fn create_new_event(user: &mut GooseUser) -> TransactionResult {
+    if !should_request_events() {
+        tokio::time::sleep(Duration::from_millis(500)).await;
```
nathanielc, line 130: Am I correct in guessing this is so we do not spin the CPU doing nothing for the node that is not writing data? If so could you add a comment to that effect?

runner/src/scenario/ceramic/event_id_sync.rs
```diff
@@ -0,0 +1,182 @@
...
+        Ok(())
+    } else {
+        let user_data: &ModelReuseLoadTestUserData = user
+            .get_session_data()
+            .expect("we are missing sync_event_id user data");
+
+        // eventId needs to be a multibase encoded string for the API to accept it
+        let event_id = format!("F{}", random_event_id(&user_data.model_id.to_string()));
+        let event_key_body = serde_json::json!({"eventId": event_id});
```
nathanielc, line 139: When my branch merges we will need to update this to have `eventData` as well which will be multibase encoded random bytes. ~1KB

nathanielc (approved):
Looks great! Just a few small comments

runner/Cargo.toml
```diff
@@ -27,3 +28,6 @@ tokio.workspace = true
 tracing-log.workspace = true
 tracing.workspace = true
 multibase.workspace = true
+
+[dev-dependencies]
+tracing-test = "0.2"
```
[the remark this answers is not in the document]
dav1do: Nice, I adjusted these tests/imports to use test_log

runner/src/scenario/ceramic/event_id_sync.rs
dav1do: Ah, good to know. Changed to 0. I wonder if this is was causing my lack of results.

runner/src/scenario/ceramic/event_id_sync.rs
dav1do: Yep, added a comment.

runner/src/scenario/ceramic/event_id_sync.rs
dav1do: Makes sense, I'll try running a test with that branch and making those updates in the meantime.

dav1do: Updated with new scenario `recon-event-sync` to include data. I also renamed the original scenario `recon-event-key-sync`.

PR #127, opened 2024-01-10
feat: Remove transaction delays to improve throughput and change meaning of scenario runner users key

dav1do:
This makes some changes to the runner to improve throughput and make the "users" key easier to understand and more likely to work.

- The `user` value in the spec is now defined as users/worker (node), rather than being the total users and needing to be divisible by the number of nodes.
- I removed the sleep between every transaction in a number of scenarios. As we're targeting one write/3.5 milliseconds, the current values aren't super insightful (way too lazy). The `SIMULATE_THROTTLE_REQUESTS` env var can be used to avoid each worker going bananas. If we want, we can expose the min/max wait as CLI options, as it can be applied to the scenario struct when building it.
- I added recording the ~~transaction, scenario, and~~ error logs on the goose workers as the scenario and transaction were added in #124.

### DIFF
--- a/runner/src/simulate.rs
+++ b/runner/src/simulate.rs
@@ -35,8 +35,14 @@ pub struct Opts {
     #[arg(long, env = "SIMULATE_PEERS_PATH")]
     peers: PathBuf,
 
-    /// Number of users to simulate
-    #[arg(long, default_value_t = 100, env = "SIMULATE_USERS")]
+    /// Number of users to simulate on each node. The total number of users 
+    /// running the test scenario will be this value * N nodes. 
+    ///
+    /// Implmentation details: A user corresponds to a tokio task responsible 
+    /// for making requests. They should have low memory overhead, so you can 
+    /// create many users and then use `throttle_requests` to constrain the overall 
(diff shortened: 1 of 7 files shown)


3benbox (approved):
LGTM

smrz2001 (approved):
🚀

runner/src/simulate.rs
```diff
@@ -206,6 +211,9 @@ fn worker_config(target_peer_addr: String, throttle_requests: Option<usize>) ->
     config.scenario_log = "scenario.log".to_owned();
     config.transaction_log = "transaction.log".to_owned();
     config.request_log = "request.log".to_owned();
+    config.transaction_log = "transaction.log".to_owned();
+    config.scenario_log = "scenario.log".to_owned();
```
nathanielc, line 215: Some of these look like duplicate lines.

dav1do: Doh, good catch. I was behind main (without realizing it somehow) and didn't notice you added it with the tokio-console changes... and apparently I botched the merge 😬.

PR #128, opened 2024-01-16
fix: only specify pubsub topic for inmemory/local

smrz2001:


operator/src/network/ceramic.rs
```diff
@@ -327,6 +322,17 @@ pub fn stateful_set_spec(ns: &str, bundle: &CeramicBundle<'_>) -> StatefulSetSpe
         },
     ];
 
+    // Only specify the pubsub topic if the network was "inmemory" or "local"
```
nathanielc, line 325: Can you update this comment for why this is important? Does js-ceramic break if we specify it for some other reason?

nathanielc (approved):
LGTM with one comment below

operator/src/network/ceramic.rs
smrz2001: Yes, js-ceramic won't come up if the pubsub topic is specified for any network type but inmemory/local.

PR #131, opened 2024-01-19
feat: add scripts to help spin up local cluster

dav1do:
I found myself nuking and re-creating my cluster _a lot_. Without a script like this, it takes forever to re-download all the images and get the cluster reconciled. This makes it much faster. You can just run `kind delete cluster && ./local-scripts/setup.sh "network.yaml"` and it will go through the steps of creating the cluster, applying the network and CRDs and then loading the images. 

It isn't one size fits all, but it does seem like a useful thing to include. If others disagree, I'm okay with closing this PR. I have a local version of .gitignore with my own folders commented with things like results, network and simulation definitions, and helper scripts. Also open to updating docs/readme etc to make the usage of these more obvious, but they're pretty straightforward scripts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,7 @@ target/
 .env*
 /book
 *.idea
+__pycache__
+
+# allow a sample network in the repo
+!local-scripts/small.yaml
--- a/local-scripts/load.py
+++ b/local-scripts/load.py
@@ -1,0 +1,45 @@
+import subprocess
+
+images = [
+        'amazon/aws-cli',
+        'ceramicnetwork/ceramic-anchor-service:latest',
+        'ceramicnetwork/composedb:latest',
+        'gresau/localstack-persist:2',
+        'public.ecr.aws/r5b3e0r5/3box/cas-contract',
+        'public.ecr.aws/r5b3e0r5/3box/ceramic-one',
+        'public.ecr.aws/r5b3e0r5/3box/go-cas:latest',
+        'trufflesuite/ganache'
+        ]
+
+
+def run_command_for_images(command: list):
+    results = []
+    for image in images:
+        exec = command.copy()
+        exec.append(image)
+        process = subprocess.Popen(exec, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        results.append(process)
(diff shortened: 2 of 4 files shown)


nathanielc (commented):
LGTM. However I am curious why you need to recreate the entire cluster vs simply recreate the network? I find I `kubectl delete network ...`  all the time but rarely delete my kind cluster. What does deleting the entire kind cluster do that deleting the network does not?

dav1do:
> LGTM. However I am curious why you need to recreate the entire cluster vs simply recreate the network? I find I `kubectl delete network ...` all the time but rarely delete my kind cluster. What does deleting the entire kind cluster do that deleting the network does not?

I need try it again and update this accordingly. It was taking take quite a while to remove the network when I had a bunch of nodes where deleting the cluster was faster (obviously coming back up wasn't). And at least once, it seemed that re-applying a different network never reconciled and got back to a ready state. It was early on in my Keramik experience, so it's likely I did something silly and/or didn't understand what I was looking at. I starting doing this when creating new topologies to make sure things were totally refreshed, especially as I was making operator and runner changes, but now that I'm not a total n00b it's probably the slowest way to go. Especially since deleting and reapplying a network is a required feature, so if it's not working right it needs to be addressed.

3benbox:
> > LGTM. However I am curious why you need to recreate the entire cluster vs simply recreate the network? I find I `kubectl delete network ...` all the time but rarely delete my kind cluster. What does deleting the entire kind cluster do that deleting the network does not?
> 
> I need try it again and update this accordingly. It was taking take quite a while to remove the network when I had a bunch of nodes where deleting the cluster was faster (obviously coming back up wasn't). And at least once, it seemed that re-applying a different network never reconciled and got back to a ready state. 

My experience has been it takes a while to delete the network, I think it's the js-ceramic container takes so long to stop, but I haven't looked into it that far.

After deleting a network, I'll do a `kubectl get ns -w` and wait for the namespace to be fully removed before redeploying the network to avoid thrashing. This has worked fine for me but does take  a little while.

3benbox (approved):
LGTM

PR #132, opened 2024-01-19
feat: add ability to disable local monitoring simulation resources

3benbox:
For the case when a keramik network is externally monitored, like on cloud infra, we should not deploy simulation monitoring resources to the namespace.

We should also allow the configuring of additional monitoring configuration through the `.spec.monitoring` path.

Works fine on my local tests, but how do I default it to true?

### DIFF
--- a/operator/src/simulation/spec.rs
+++ b/operator/src/simulation/spec.rs
@@ -31,6 +31,24 @@ pub struct SimulationSpec {
     /// validate throughput and correctness before returning. The exact definition is
     /// left to the scenario (requests per second, total requests, rps/node etc).
     pub success_request_target: Option<usize>,
+    /// Monitoring resources for a simulation
+    pub monitoring: Option<MonitoringSpec>,
+
+}
+/// MonitoringSpec defines how Simulation will be monitored.
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct MonitoringSpec {
+    /// Monitoring in the namespace
+    pub local: Option<LocalMonitoringSpec>,
+}
+
+/// LocalMonitoringSpec defines monitoring in the namespace.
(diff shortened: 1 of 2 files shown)


operator/src/simulation/spec.rs
```diff
@@ -31,6 +31,24 @@ pub struct SimulationSpec {
     /// validate throughput and correctness before returning. The exact definition is
     /// left to the scenario (requests per second, total requests, rps/node etc).
     pub success_request_target: Option<usize>,
+    /// Monitoring resources for a simulation
+    pub monitoring: Option<MonitoringSpec>,
```
dav1do, line 35: My suggestion would be to use serde default and convert this to an enum.. for example:

```rust
    #[serde(default)]
    pub monitoring: MonitoringSpec,
}
/// MonitoringSpec defines how Simulation will be monitored.
#[derive(Serialize, Deserialize, Debug, PartialEq, Clone, JsonSchema)]
#[serde(rename_all = "camelCase")]
pub enum MonitoringSpec {
    /// Monitoring in the namespace
    Local(LocalMonitoringSpec),
}

impl Default for MonitoringSpec {
    fn default() -> Self {
        MonitoringSpec::Local(LocalMonitoringSpec::enabled())
    }
}

/// LocalMonitoringSpec defines monitoring in the namespace.
#[derive(Serialize, Deserialize, Debug, PartialEq, Clone, JsonSchema)]
#[serde(rename_all = "camelCase")]
pub struct LocalMonitoringSpec {
    /// Deploy monitoring to the namespace
    pub enabled: bool,
}

impl LocalMonitoringSpec {
    /// Create a new LocalMonitoringSpec with enabled set to true
    pub fn enabled() -> Self {
        LocalMonitoringSpec { enabled: true }
    }
}
```

[the serde docs](https://serde.rs/attr-default.html) have more info about options for specifying/generating defaults

3benbox:
no longer needed

PR #133, opened 2024-01-24
chore: update dependencies (primarily k8s related)

dav1do:
The k8s API layout changed and I had to regenerate the expect results, but the data looks the same (just key order changes -> they're sorted now!). I deployed a network and simulation (the recon sync test using redis) and it all seemed to work well.


smrz2001 (approved):
Thanks for taking care of this! 🚀 

Just one question - should we bump up the operator [version](https://github.com/3box/keramik/pull/133/files#diff-fa2e3014b632401a214e3fd555ac30a556df7ca84e01192da5344c8fc8b34b57R4)?

I don't have a preference, just wondering if that's the way it should be done. That's the piece that gets [used](https://github.com/3box/ceramic-tests/blob/main/hermetic/Cargo.toml#L23) by the `ceramic-tests` hermetic driver.

dav1do:
> Thanks for taking care of this! 🚀
> 
> Just one question - should we bump up the operator [version](https://github.com/3box/keramik/pull/133/files#diff-fa2e3014b632401a214e3fd555ac30a556df7ca84e01192da5344c8fc8b34b57R4)?
> 
> I don't have a preference, just wondering if that's the way it should be done. That's the piece that gets [used](https://github.com/3box/ceramic-tests/blob/main/hermetic/Cargo.toml#L23) by the `ceramic-tests` hermetic driver.

Hm, that's a good question. We probably should start versioning this, especially as we're depending on it elsewhere. I'm happy to bump the version (0.1.0, 0.0.2?), but unless we specify a version key in ceramic-tests, it won't really apply. 

That said, versions with git dependencies are weird. The [cargo docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories) say this:
>When the dependency is retrieved from git, the version key will not affect which commit is used, but the version information in the dependency’s Cargo.toml file will still be validated against the version requirement.

so basically, when you `cargo update` it will update to the newest main commit as long the version is compatible. Or maybe it will update and then fail? I should test it...

smrz2001:
> > Thanks for taking care of this! 🚀
> > Just one question - should we bump up the operator [version](https://github.com/3box/keramik/pull/133/files#diff-fa2e3014b632401a214e3fd555ac30a556df7ca84e01192da5344c8fc8b34b57R4)?
> > I don't have a preference, just wondering if that's the way it should be done. That's the piece that gets [used](https://github.com/3box/ceramic-tests/blob/main/hermetic/Cargo.toml?rgh-link-date=2024-01-25T00%3A26%3A18Z#L23) by the `ceramic-tests` hermetic driver.
> 
> Hm, that's a good question. We probably should start versioning this, especially as we're depending on it elsewhere. I'm happy to bump the version (0.1.0, 0.0.2?), but unless we specify a version key in ceramic-tests, it won't really apply.
> 
> That said, versions with git dependencies are weird. The [cargo docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories) say this:
> 
> > When the dependency is retrieved from git, the version key will not affect which commit is used, but the version information in the dependency’s Cargo.toml file will still be validated against the version requirement.
> 
> so basically, when you `cargo update` it will update to the newest main commit as long the version is compatible. Or maybe it will update and then fail? I should test it...

What would you recommend? I'd be cool either way.

dav1do:
> > > Thanks for taking care of this! 🚀
> > > Just one question - should we bump up the operator [version](https://github.com/3box/keramik/pull/133/files#diff-fa2e3014b632401a214e3fd555ac30a556df7ca84e01192da5344c8fc8b34b57R4)?
> > > I don't have a preference, just wondering if that's the way it should be done. That's the piece that gets [used](https://github.com/3box/ceramic-tests/blob/main/hermetic/Cargo.toml?rgh-link-date=2024-01-25T00%3A26%3A18Z#L23) by the `ceramic-tests` hermetic driver.
> > 
> > 
> > Hm, that's a good question. We probably should start versioning this, especially as we're depending on it elsewhere. I'm happy to bump the version (0.1.0, 0.0.2?), but unless we specify a version key in ceramic-tests, it won't really apply.
> > That said, versions with git dependencies are weird. The [cargo docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories) say this:
> > > When the dependency is retrieved from git, the version key will not affect which commit is used, but the version information in the dependency’s Cargo.toml file will still be validated against the version requirement.
> > 
> > 
> > so basically, when you `cargo update` it will update to the newest main commit as long the version is compatible. Or maybe it will update and then fail? I should test it...
> 
> What would you recommend? I'd be cool either way.

I guess my opinion is that it's not important to manage the version until we're downloading the project as a crate from somewhere, or it's being used as a library in another library (i.e. no Cargo.lock). Right now, the lock file pins the commit hash, so you have to go in and cargo update to get the new version of main. 

If we added a version, we'd get errors like this if it's not met:

```sh
❯ cargo update -p keramik-operator
    Updating git repository `https://github.com/3box/keramik.git`
    Updating crates.io index
error: failed to select a version for the requirement `keramik-operator = "^0.0.2"`
candidate versions found which didn't match: 0.0.1
location searched: Git repository https://github.com/3box/keramik.git?branch=main
required by package `hermetic-driver v0.1.0 (/Users/david/3box/ceramic-tests/hermetic)`
```

so we could add `version = "0.0"` and then cargo update should keep working as long as we don't implement a breaking (minor) update, which for <1.0 crates minor is usually considered breaking. Going to ask in the keramik channel. My opinion is it's not worth it until we iron out our current CI flow around keramik -> ceramic-tests but we're getting close to that.

smrz2001:
> so we could add `version = "0.0"` and then cargo update should keep working as long as we don't implement a breaking (minor) update, which for <1.0 crates minor is usually considered breaking. Going to ask in the keramik channel. My opinion is it's not worth it until we iron out our current CI flow around keramik -> ceramic-tests but we're getting close to that.

Yes, I'm cool with waiting till later 👍🏼

PR #135, opened 2024-01-29
feat: move monitoring resources to network

nathanielc:
There is now a `monitoring` flag on networks when true deploys opentelemetry, jaeger and prometheus. Opentelemetry can been changed to no longer scrape targets, rather it only listens for pushed metrics.

The simulation no longer deploys any of these resources. However the simulation does check that they exist and are running in order to start a simulation. This means that on our network were we intend to run simulations we need to set `monitoring=true` on the network.

Additionally logic was added to the simulate logic inside the runner to report the minimum rps of any peer after a simulation completes.

With these two changes we should be able to capture the min_peer_rps as the most important metrics from simulation runs and persist that metric in our cloud systems.


runner/src/simulate.rs
```diff
@@ -628,17 +654,23 @@ impl Metrics {
...
             .lock()
             .expect("should be able to acquire metrics lock for mutation");
         gm.goose_metrics = Some(metrics);
+        gm.min_peer_rps = min_peer_rps;
     }
 }
 
 impl MetricsInner {
     fn observe(&mut self, cx: &Context) {
+        // TODO add simulation specific attributes
```
dav1do, line 669: was going to comment on this, but never mind for now. we should be able to keep going like this but it'd be nice to have some label to make sure we're comparing the same thing over time

dav1do (commented):
LGTM. Mind updating the docs so this new setting is included in the examples? It's pretty essential to running a simulation and would be weird if it didn't exist (we could have the simulation deploy those resources too.. but that seems messy). 

I also wanted to check in make sure we don't need to start versioning operator or anything. I'm pretty sure it's fine since we just pin SHA in ceramic-tests, but I'm not sure how all the CI stuff is updated. I want to make sure changes aren't too disruptive as we need to update those networks to include the monitoring flag now (and I think my changes to add the new bool for devMode caused some sim/network/keramik mismatches that would be nice to avoid if possible).

3benbox:
I'd really like the `monitoring` option to be able to handle multiple monitoring situations.

Can we make this an object like this as the default?
```
spec:
  monitoring:
    local: true
```

The idea being future monitoring can be added like:
```
spec:
  monitoring:
    local: false
    gcpPodMonitoring: true
    grafanaCloud:
      secretName:
      secretAPIKey:
```

I don't care what "local" or "namespaced" monitoring key is, but something like that.

nathanielc:
@dav1do @3benbox Thanks for the feedback. I have updated the book and added a few features too.

1. We generate a random name for each simulation run and tag the metrics with that name
2. The monitoring field is a struct that we can extend with more options in the future
3. This _should_ not be a breaking change as it only adds optional items to the spec.

To upgrade we should:

1. Wait for any simulations to finish, then delete then
2. Update the CRD definitions
3. Restart the operator
4. Update the networks to add `monitoring: {namespaced: true }` and apply

dav1do (approved):

3benbox:
Amazing! 🤩

PR #138, opened 2024-02-01
feat: expose prometheus metrics from operator

nathanielc:
Now the operator exposes a prometheus scrape endpoint. It does not try to push its metrics over OTLP. This operator is long lived so this should not be a problem.

The runner continues to push its metrics over OTLP which is important since it is short lived.

However I was not able to get both prometheus scrape and OTLP push working on the same process. So the operator only supports prometheus and the runner only support OTLP push.

### DIFF
--- a/common/src/telemetry.rs
+++ b/common/src/telemetry.rs
@@ -1,19 +1,31 @@
 //! Provides helper functions for initializing telemetry collection and publication.
-use std::time::Duration;
+use std::{convert::Infallible, net::SocketAddr, sync::OnceLock, time::Duration};
 
 use anyhow::Result;
-use opentelemetry::{
-    runtime,
-    sdk::{
-        export::metrics::aggregation,
-        metrics::{controllers::BasicController, selectors},
-    },
+use hyper::{
+    body::Body,
+    http::HeaderValue,
+    service::{make_service_fn, service_fn},
+    Request, Response,
 };
+use opentelemetry::global;
 use opentelemetry_otlp::WithExportConfig;
+use opentelemetry_sdk::{
+    metrics::{
+        reader::{DefaultAggregationSelector, DefaultTemporalitySelector},
+        MeterProvider,
+    },
+    runtime, trace, Resource,
+};
+use prometheus::{Encoder, TextEncoder};
+use tokio::{sync::oneshot, task::JoinHandle};
 use tracing_subscriber::{filter::LevelFilter, prelude::*, EnvFilter, Registry};
 
-/// Initialize tracing and metrics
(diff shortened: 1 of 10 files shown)


operator/src/network/controller.rs
```diff
@@ -202,6 +203,13 @@ async fn reconcile(
     network: Arc<Network>,
     cx: Arc<Context<impl IpfsRpcClient, impl RngCore, impl Clock>>,
 ) -> Result<Action, Error> {
+    let meter = global::meter("keramik");
+    let runs = meter
+        .u64_counter("network_reconcile_count")
```
3benbox, line 208: Can we also get an error count? 
When the operator is unable to reconcile a network, increase a counter?

operator/src/simulation/controller.rs
```diff
@@ -128,6 +129,13 @@ async fn reconcile(
     simulation: Arc<Simulation>,
     cx: Arc<Context<impl IpfsRpcClient, impl RngCore, impl Clock>>,
 ) -> Result<Action, Error> {
+    let meter = global::meter("keramik");
+    let runs = meter
+        .u64_counter("simulation_reconcile_count")
```
3benbox, line 134: Same with simulation failures.

runner/src/simulate.rs
```diff
@@ -635,6 +639,19 @@ impl Metrics {
...
+            users_total.as_any(),
+            scenarios_total.as_any(),
+            scenarios_duration_percentiles.as_any(),
+            txs_total.as_any(),
+            txs_duration_percentiles.as_any(),
+            requests_total.as_any(),
+            requests_status_codes_total.as_any(),
+            requests_duration_percentiles.as_any(),
+            simulation_min_peer_requests_per_second.as_any(),
+        ];
```
3benbox, line 654: 🤩

common/src/telemetry.rs
```diff
@@ -90,5 +74,93 @@ pub async fn init(otlp_endpoint: String) -> Result<BasicController> {
...
+        .set(prom_registry)
+        .expect("should be able to set PROM_REGISTRY");
+    let meter = MeterProvider::builder().with_reader(prom_exporter).build();
+    global::set_meter_provider(meter.clone());
+
+    let (shutdown, join) = start_metrics_server(addr);
+    Ok((meter, shutdown, join))
+}
+
+// /metrics scrape endpoin
```
3benbox, line 130: comment spelling, "endpoint" 😇

3benbox (changes requested):
Amazing, I'd just like to see some error metrics when things don't go as expected.

Added comments.

3benbox (approved):
LGTM

PR #139, opened 2024-02-06
feat: add podmonitor crd for keramik operator deployments

3benbox:
Adds a podmonitor CRD to capture keramik operator metrics.

Deployments would add this resource to their kustomization.yaml to enable.


smrz2001 (approved):
🚀

PR #140, opened 2024-02-06
fix: use labels instead of hardcoded indexes

nathanielc:
This PR makes two changes.

First we use labels to track which stateful sets and services should belong to a ceramic configuration. This changes from the previous logic that hardcoded indexes and deleted resources blindly based on those indexes. This reduces the noise in the operator logs as it only does the work that is needed vs always assuming there are resources to delete.

Second this changes both network and simulations to not watch nearly as many resources. This greatly reduces the number of reconcile loops we run and thus the amount of work the operator needs to do. And we still get low latency when a network or simulation is changed.

These changes combined should improve the stability of the keramik operator.


smrz2001 (approved):
LGTM! :rocket:

PR #145, opened 2024-02-09
refactor: small change to only count connected peers

nathanielc:
When reviewing the #144 I noticed a small optimization we can apply to the logic. Instead of collecting the peers into a vector and then using the length of the vector we can simply count the iterator directly, thus avoiding the allocations of the vector. Not that this code is perf sensitive but generally improves the readability of the code as its intention is clear.


smrz2001 (approved):
Ty! 🥳

PR #146, opened 2024-02-09
fix: add n/w check for cas ipfs during peer status updates

smrz2001:
CAS IPFS status lookup for durable networks pointing to external CAS is causing peer status reconciliation to fail.

I've only tested this on the Clay/Prod cluster so far, but their tests are now green!

https://github.com/3box/ceramic-tests/actions/runs/7848110468
https://github.com/3box/ceramic-tests/actions/runs/7848846872

### DIFF
--- a/operator/src/network/ipfs_rpc.rs
+++ b/operator/src/network/ipfs_rpc.rs
@@ -14,7 +14,7 @@ pub trait IpfsRpcClient {
 /// Information about connected peers
 #[derive(Debug, Clone, serde::Deserialize)]
 pub struct Peer {
-    pub addr: String,
+    #[serde(rename = "Peer")]
     pub id: String,
 }
 
--- a/operator/src/network/controller.rs
+++ b/operator/src/network/controller.rs
@@ -375,7 +375,7 @@ async fn reconcile_(
         delete_service(cx.clone(), &ns, &service.name_any()).await?;
     }
 
-    // Reconile replica counts
+    // Reconcile replica counts
     let computed_replicas = ceramics
         .iter()
         .fold(0, |acc, bundle| acc + bundle.info.replicas);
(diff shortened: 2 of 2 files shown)


operator/src/network/ipfs_rpc.rs
```diff
@@ -14,7 +14,7 @@ pub trait IpfsRpcClient {
 /// Information about connected peers
 #[derive(Debug, Clone, serde::Deserialize)]
 pub struct Peer {
-    pub addr: String,
+    #[serde(rename = "Peer")]
```
dav1do, line 17: What's the deal here? I see `Peer` as a field in [the kubo docs](https://docs.ipfs.tech/reference/kubo/rpc/#api-v0-swarm-peers) but I'm confused about why this was needed (or how this was fine before since it seems like it should have been `Id` and `Addr`).

dav1do (approved):
Looks good! I added a question about the change to the Peer struct. It'd be nice if we could somehow reconcile with a list of expected resources, rather than have to check conditions in operator and controller to decide what should exist, but that might not make it any simpler... just wondering out loud 😁

operator/src/network/ipfs_rpc.rs
smrz2001: Yeah, I can see how that would be confusing haha.

Before the previous PR for filtering on Keramik peer IDs, we actually didn't care about _any_ of the fields in that struct:
```
        #[derive(serde::Deserialize)]
        struct Peer {
            //We currently do not care about any of the information of connected peers,
            // just the number of them. As a result this struct does not have any fields.
        }
        #[derive(serde::Deserialize)]
        struct Response {
            #[serde(rename = "Peers")]
            peers: Option<Vec<Peer>>,
        }
```
Then the last PR started caring about the fields:
```
/// Information about connected peers
#[derive(Debug, Clone, serde::Deserialize)]
pub struct Peer {
    pub addr: String,
    pub id: String,
}
```
However, there was a bug in my code. I changed the field name to `id`, when in fact, the API returns `peer` 🙄

With this latest change, we removed `addr` because we don't really care about it, and are deserializing the peer ID correctly. I think `id` is clearer than `peer`.

smrz2001:
> Looks good! I added a question about the change to the Peer struct. It'd be nice if we could somehow reconcile with a list of expected resources, rather than have to check conditions in operator and controller to decide what should exist, but that might not make it any simpler... just wondering out loud 😁

Yeah, let me know if you want to chat about this more next week! I'll merge this PR to get the operator working again but we can chat about further improvements.

operator/src/network/ipfs_rpc.rs
dav1do: Oh right, that makes sense. We used to just count elements in the list and now we actually check the id. Thanks for explaining!

PR #147, opened 2024-02-10
chore: update to use new create event payload

dav1do:
Update recon create event request to use new [payload format](https://github.com/ceramicnetwork/rust-ceramic/pull/269)


smrz2001 (approved):
:shipit:

PR #148, opened 2024-02-13
feat: expose storage class for IPFS containers

nathanielc:


operator/src/network/ipfs.rs
```diff
@@ -110,6 +117,7 @@ impl Default for RustIpfsConfig {
                 memory: Some(Quantity("512Mi".to_owned())),
                 storage: Quantity("1Gi".to_owned()),
             },
+            storage_class: None,
```
3benbox, line 120: I'm not sure exactly how this will evaluate with Rust.

If there is no storageClassName specified, then the PVC should not set one, not even an empty string, so that the default storageClassName is applied.

https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1

3benbox (commented):
A question about defaults

operator/src/network/ipfs.rs
nathanielc: That is what is happening, if no storage class is specified in the spec then its not set on the PVC spec itself. Rust does this via the `Option<String>` type, where `None` means its not set and `Some("some_storage_class_name")` is set.

3benbox (approved):
LGTM

PR #150, opened 2024-02-14
feat: adds node-affinity spec config

nathanielc:
When node_selector_terms is set on a network all pods in the network will use those terms in order to control on which nodes the networks pods are scheduled.

This does not affect pods in simulations. I can add that if its needed/wanted.


3benbox (approved):
Deployed and tested on github-actions cluster

PR #152, opened 2024-02-16
chore: use interests instead of subscribe recon endpoint in simulation

dav1do:
Use interests endpoint instead of deprecated subscribe endpoint for recon simulation.


smrz2001 (approved):
🚀

PR #153, opened 2024-02-16
fix: no longer reuse the secret from cas for ceramic

nathanielc:
If CAS was not configured then ceramic postgres would not have a secret. Now they each have their own secret.


smrz2001 (approved):
🥳

3benbox (approved):
LGMT

PR #154, opened 2024-02-16
fix: no default otel endpoint for the operator

nathanielc:
The operator no longer has a default opentelemetry endpoint and will only collect and send traces if an endpoint is configured.

This might address the memory leak but either way its a nice quality of life improvement as we have not been collecting traces from the operator anyway.

### DIFF
--- a/k8s/operator/manifests/operator.yaml
+++ b/k8s/operator/manifests/operator.yaml
@@ -127,9 +127,6 @@ spec:
           containerPort: 9464
           protocol: TCP
         env:
-        # We are pointing to tempo or grafana tracing agent's otlp grpc receiver port
-        - name: OPERATOR_OTLP_ENDPOINT
-          value: "https://otel:4317"
         - name: RUST_LOG
           value: "info"
         #readinessProbe:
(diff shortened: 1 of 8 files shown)


3benbox (approved):
lgtm

nathanielc:
Turns out this breaks logging within the keramik operator. Still trying to figure out why.

smrz2001 (approved):
🚀

k8s/operator/manifests/operator.yaml
```diff
@@ -127,9 +127,6 @@ spec:
           containerPort: 9464
           protocol: TCP
         env:
-        # We are pointing to tempo or grafana tracing agent's otlp grpc receiver port
-        - name: OPERATOR_OTLP_ENDPOINT
-          value: "https://otel:4317"
```
nathanielc, line 132: We need to remove this config option from any operators we have configured otherwise they will still try to send traces.

nathanielc:
@3benbox @smrz2001 I could use another review. The second commit fixes the issue with no longer logging any events to stdout. 

 Local testing shows this fixed the OOM issues. The traces must have been being cached forever since they could not be flushed. 

In order to fix this in our prod envs we will need to remove the OTLP_ENDPOINT env var from the deployment of the operator.

PR #156, opened 2024-02-25
feat: pod monitoring

smrz2001:
This PR adds automated creation of pod monitoring custom resources that can be used by the Grafana Cloud Agent for Keramik networks.

### DIFF
--- a/keramik/src/monitoring.md
+++ b/keramik/src/monitoring.md
@@ -44,7 +44,14 @@ Scrape the `otel:9465` endpoint if you want on the simulation metrics.
 
 ## Pod Monitoring
 
-If pod monitoring is enabled, the operator will create custom resources for monitoring the network pods. The Grafana
-Cloud Agent will then consume this configuration and send the metrics to Grafana Cloud.
-
 This option expects the `PodMonitor` custom resource definition to already be installed in the network namespace.
+
+If `podMonitor` is enabled, the operator will create `podmonitors.monitoring.coreos.com` resources for collecting the metrics from the pods in the network.
+
+If you're using something like the grafana cloud agent, or prometheus-operator, the `podmonitors.monitoring.coreos.com` will be installed already.
+
+You can install the CRD directly from the operator:
+
+```
+    kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/main/example/prometheus-operator-crd/monitoring.coreos.com_podmonitors.yaml
+```
--- a/keramik/src/monitoring.md
+++ b/keramik/src/monitoring.md
@@ -18,6 +18,7 @@ spec:
   replicas: 2
   monitoring:
     namespaced: true
+    podMonitor: true
 ```
 
 To view the metrics and traces port-forward the services:
@@ -40,3 +41,10 @@ This is typically a collection of metrics per simulation run and is much lighter
 Scrape the `otel:9465` endpoint if you want on the simulation metrics.
 
 >NOTE: The prometheus-0 pod will scrape all metrics so you can easily inspect all activity on the network.
+
+## Pod Monitoring
+
+If pod monitoring is enabled, the operator will create custom resources for monitoring the network pods. The Grafana
+Cloud Agent will then consume this configuration and send the metrics to Grafana Cloud.
+
+This option expects the `PodMonitor` custom resource definition to already be installed in the network namespace.
--- a/operator/src/network/controller.rs
+++ b/operator/src/network/controller.rs
@@ -14,15 +14,16 @@ use k8s_openapi::{
         batch::v1::Job,
         core::v1::{Namespace, Pod, Secret, Service, ServiceStatus},
     },
+    apiextensions_apiserver::pkg::apis::apiextensions::v1::CustomResourceDefinition,
     apimachinery::pkg::apis::meta::v1::Time,
 };
 use keramik_common::peer_info::{CeramicPeerInfo, Peer, PeerId};
 use kube::{
-    api::{DeleteParams, ListParams, Patch, PatchParams},
+    api::{DeleteParams, ListParams, Patch, PatchParams, PostParams},
     client::Client,
     core::{object::HasSpec, ObjectMeta},
     runtime::Controller,
-    Api, ResourceExt,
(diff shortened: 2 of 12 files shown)


operator/src/network/controller.rs
```diff
@@ -963,6 +973,54 @@ async fn reset_bootstrap_job(
...
+                        target_port: Some(monitor_port),
+                    }],
+                    selector: Some(SelectorSpec {
+                        match_labels: {
+                            let mut map = BTreeMap::new();
+                            map.insert("app".to_owned(), monitor_label.to_owned());
+                            map
+                        },
+                    }),
+                },
```
3benbox, line 1007: We should add another section to the podmonitor,   `podTargetLabels`

If should be a list of strings for pod labels that should be propagated to the metrics as labels.

The list could come from the Network definition, or the static list I'm using:
```
  podTargetLabels:
    - affinity
    - scenario
    - simulation
  ```

**Review by 3benbox — CHANGES_REQUESTED** [review]
looks good, but   podTargetLabels should be included for mapping metrics to the Simulation.

File: operator/src/network/controller.rs (hunk 1 above)
Comment (smrz2001) [reply operator/src/network/controller.rs:1007]: Ah yes, thanks for reminding me. I thought I was forgetting something 😊

File: operator/src/network/spec.rs [hunk 2]
```diff
@@ -265,4 +265,56 @@ pub struct ResourceLimitsSpec {
...
+    /// Deploy pod monitors
+    pub pod_monitoring: Option<PodMonitoringSpec>,
+}
+
+/// Describes the pod monitor configuration
+#[derive(Serialize, Deserialize, Debug, Default, PartialEq, Clone, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct PodMonitoringSpec {
+    /// Whether pod monitoring is enabled
+    pub enabled: Option<bool>,
```
dav1do, line 277: since the parent level is optional, is there a reason to also make this field optional?

keramik/src/monitoring.md
```diff
@@ -18,6 +18,7 @@ spec:
   replicas: 2
   monitoring:
     namespaced: true
+    podMonitor: true
```
dav1do, line 21:
```suggestion
    podMonitoring:
      enabled: true
```

operator/src/network/spec.rs
dav1do: err.. maybe nevermind given these.. as this is what the user interacts with.

```rust
#[derive(Default)]
struct MonitoringConfig {
    namespaced: bool,
    pod_monitoring: PodMonitoringConfig,
}

#[derive(Default)]
struct PodMonitoringConfig {
    enabled: bool,
    pod_target_labels: Vec<String>,
}
```

keramik/src/monitoring.md
```diff
@@ -40,3 +41,10 @@ This is typically a collection of metrics per simulation run and is much lighter
 Scrape the `otel:9465` endpoint if you want on the simulation metrics.
 
 >NOTE: The prometheus-0 pod will scrape all metrics so you can easily inspect all activity on the network.
+
+## Pod Monitoring
+
+If pod monitoring is enabled, the operator will create custom resources for monitoring the network pods. The Grafana
+Cloud Agent will then consume this configuration and send the metrics to Grafana Cloud.
+
+This option expects the `PodMonitor` custom resource definition to already be installed in the network namespace.
```
dav1do, line 50: how do you install this CRD? I feel like I'm missing something... locally, I apply a grafana deployment.. would this work work similarly or is this purely for cloud/durable environments?

dav1do (approved):
A couple comments, but overall looks okay to me. I can't say I totally understand it though 😅

keramik/src/monitoring.md
3benbox: A bit of a mixup, these should ALL be `podmonitor` and no `podmonitoring` at all.

Confusingly, `podmonitor` is part of an upstream prometheus operator project that a few projects use and the Grafana Cloud agent supports.
https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#monitoring.coreos.com/v1.PodMonitor

`podmonitoring` is a GCP provided, prometheus compatible, cluster provided, CRD that we where thinking about using.
https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/doc/api.md#monitoring.googleapis.com/v1.PodMonitoring

keramik/src/monitoring.md
3benbox: The upstream CRD is from the prometheus-operator project above, but lots of projects redistribute it including the grafana agent installation https://grafana.com/docs/agent/latest/operator/#static-mode-kubernetes-operator-beta

Functionally I was thinking we should check if the CRD is installed, because otherwise the CRD instance deployment will fail.

But do not install it if it's not there (redistribute) because that would mean no one is monitoring the CRD instance, so why waste resources.

operator/src/network/spec.rs
smrz2001: Yeah, I had the same question. We decided to do it this way to mimic the general pattern for how k8s resources are configured, i.e. they can be configured but not enabled.

keramik/src/monitoring.md
dav1do: Thanks.. that mostly makes sense 😅 I was trying to update the docs to match the spec so it actually succeeds, but it didn't do anything since the crd had to be installed. I'm good with whatever you guys decide, but if you could update the docs with how (if) it's something that should be done locally, that'd be awesome.

smrz2001: Yeah, agreed.

@3benbox, would you mind telling me how the CRD normally gets created 😅? I can add it to the README.

smrz2001: Hey @dav1do, I'll add this to the README once I get a chance to chat with Ben. I'm going to merge the PR because we're stuck in our testing without pod monitoring and we don't know how to create them quickly by hand 🙈

keramik/src/monitoring.md
3benbox: Updated the readme @dav1do , does this make sense?

3benbox (approved):
LGTM

PR #157, opened 2024-02-26
fix: expose min_peer_rps metric from CeramicNewStreamsBenchmark 

dav1do:
I calculated the value but didn't return it so that it was exported to grafana. That has been remedied.


3benbox (approved):
lgtm

PR #159, opened 2024-03-05
feat: record all peer rps in grafana for simulations to calculate avg/min/max 

dav1do:
We capture all peers average RPS over the simulation in `simulation_peer_requests_per_second` with an attribute of `peer`. We still have `simulation_min_peer_requests_per_second`. The motivation is being able to identify if changing something increases node writes, but reduces recon sync. We want a more clear picture of total writes in the scenario.

example includes the `ceramic-new-streams-benchmark` and `recon-event-sync` scenarios:
image

### DIFF
--- a/runner/src/simulate.rs
+++ b/runner/src/simulate.rs
@@ -206,33 +206,108 @@ impl MetricsCollection for PromMetricCollector {
 }
 
 #[derive(Debug, Clone)]
-struct PeerRps(Vec<f64>);
+struct PeerRequestInfo {
+    pub name: String,
+    pub count: u64,
+    pub runtime: Option<u64>,
+}
+
+impl PeerRequestInfo {
+    pub fn new(name: String, count: u64, runtime: Option<u64>) -> Self {
+        Self {
+            name,
+            count,
+            runtime,
+        }
+    }
+
+    pub fn rps(&self) -> f64 {
+        match self.runtime {
+            Some(0) => {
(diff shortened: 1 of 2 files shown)


3benbox:
What I'd rather see is just the rps per peer using labels on a single metric.
We can calculate max/min/avg from a query.

It would look something like `simulation_requests_per_second{peer="ceramic-0"}`
or maybe more accurately. `simulation_results_requests_per_second{peer="ceramic-0"}` since these are a calculation.

dav1do:
> What I'd rather see is just the rps per peer using labels on a single metric. We can calculate max/min/avg from a query.
> 
> It would look something like `simulation_requests_per_second{peer="ceramic-0"}` or maybe more accurately. `simulation_results_requests_per_second{peer="ceramic-0"}` since these are a calculation.

Reworked to include `simulation_peer_requests_per_second` with a `peer` attribute (and rebased with main).

runner/src/simulate.rs
```diff
@@ -205,6 +205,111 @@ impl MetricsCollection for PromMetricCollector {
     }
 }
 
+#[derive(Debug, Clone)]
+struct PeerRequestInfo {
```
3benbox, line 209: Just a name comment, I feel like we should have Metric in the name or something to distinguish this from Peer Requesting Info 😵 

PeerRequestMetricInfo?

3benbox (approved):
Comment about naming.
Feature is amazing!
Someone else should review for code correctness

PR #160, opened 2024-03-08
feat: add ability to set Ceramic network on simulations

nathanielc:
Prior to this change the simulations would always assume a Ceramic network of local 0. Now it is configurable.

NOTE: This is not ideal as it requires that both the network and the simulation are configured with the same Ceramic network id. However once we make the changes to remove the EventID from the Ceramic API this will no longer be necessary and can be removed.

### DIFF
--- a/keramik/src/simulation.md
+++ b/keramik/src/simulation.md
@@ -23,6 +23,7 @@ metadata:
   namespace: keramik-<unique-name>-small
 spec:
   scenario: ceramic-simple
+  network: local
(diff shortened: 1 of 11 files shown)


3benbox (changes requested):
LGTM
Can you update the docs?
We'll need to update CRDs for this change too, yeah?

nathanielc:
This is no longer needed as the clients no longer need to know what network the server is using.

PR #161, opened 2024-03-21
feat: add init image name for ceramic

3benbox:
Currently we're using the same image for the ceramic service init container that runs the actual workload.

We can't guarantee the composedb cli is available in the image for the private-key -> did transform (it's not by default).

This PR uses the new `"ceramicnetwork/composedb-cli:latest` image which has the required tooling for the init container and exposes it for other init needs.

### DIFF
--- a/operator/src/network/ceramic.rs
+++ b/operator/src/network/ceramic.rs
@@ -135,6 +135,7 @@ pub fn service_spec() -> ServiceSpec {
 pub struct CeramicConfig {
     pub weight: i32,
     pub init_config_map: String,
+    pub init_image_name: String,
     pub image: String,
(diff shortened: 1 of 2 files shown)


nathanielc (commented):
Looks good but I don't see the line of code that actually uses this new config to construct the spec. Also I would expect some tests to need to be updated as well.

smrz2001:
> Looks good but I don't see the line of code that actually uses this new config to construct the spec. Also I would expect some tests to need to be updated as well.

Yes, that part's on me, will add that today! Was going to do that yesterday but didn't get a chance.

dav1do (approved):

PR #163, opened 2024-04-19
feat: add debug mode to network

nathanielc:
This creates a new debug mode flag on the network spec that enables security contexts and tokio console as appropriate.

This will require we redeploy the crds to the clusters but is a backwards compatible change.


3benbox (approved):
lgtm

PR #164, opened 2024-04-23
Feature/cas load test scenario

samika98:
Load test that only generates load currently. We are storing the streams in redis if in future we want to perform a validation on this test. 

Currently, this test is being used generate load on cas and we monitor the dashboard to ensure CAS is not falling over. 

Since Goose is distributed load generator it is easier to generate load using this scenario compared to pocket-knife.

This scenario creates a Goose transaction which hits the CAS createAnchor API. Most of the code in anchor.rs creates an auth payload for the PUT request and a CAR file

### DIFF
--- a/runner/src/scenario/ceramic/anchor.rs
+++ b/runner/src/scenario/ceramic/anchor.rs
@@ -1,6 +1,8 @@
 use anyhow::Result;
 use base64::{engine::general_purpose, Engine};
-use ceramic_core::{Cid, DidDocument, JwkSigner, Jws, StreamId, StreamIdType};
+use ceramic_http_client::ceramic_event::{
+    Cid, DidDocument, JwkSigner, Jws, StreamId, StreamIdType,
+};
 use chrono::Utc;
 use goose::prelude::*;
 use iroh_car::{CarHeader, CarWriter};
@@ -8,12 +10,12 @@ use libipld::{cbor::DagCborCodec, ipld, prelude::Codec, Ipld, IpldCodec};
 use multihash::{Code::Sha2_256, MultihashDigest};
 use rand::{distributions::Alphanumeric, thread_rng, Rng};
 use redis::{aio::MultiplexedConnection, AsyncCommands};
-use reqwest::Client;
 use serde::{Deserialize, Serialize};
 use std::{collections::BTreeMap, sync::Arc};
+use tokio::sync::Mutex;
 use uuid::Uuid;
 
-use crate::scenario::{ceramic::model_instance::CeramicModelInstanceTestUser, get_redis_client};
+use crate::scenario::get_redis_client;
 
 #[derive(Serialize, Deserialize)]
 struct CasAuthPayload {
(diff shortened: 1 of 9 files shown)


runner/src/scenario/ceramic/anchor.rs
```diff
@@ -0,0 +1,174 @@
...
+
+pub async fn cas_benchmark() -> Result<Scenario, GooseError> {
+    let redis_cli = get_redis_client().await.unwrap();
+    let multiplexed_conn = redis_cli.get_multiplexed_tokio_connection().await.unwrap();
+    let conn_mutex = Arc::new(Mutex::new(multiplexed_conn));
+
+    let create_anchor_request = Transaction::new(Arc::new(move |user| {
+        let conn_mutex_clone = conn_mutex.clone();
+        Box::pin(async move {
+            let mut conn = conn_mutex_clone.lock().await;
```
dav1do, line 133: same comment as #162 about moving the lock internally

runner/src/scenario/ceramic/mod.rs
```diff
@@ -67,7 +68,10 @@ impl Credentials {
         let doc: DidDocument = DocumentBuilder::default()
             .id(did)
             .build()
-            .map_err(|e| anyhow::anyhow!("failed to build DID document: {}", e))?;
+            .map_err(|e| anyhow::anyhow!("failed to build DID document: {}", e))
+            .map(|core_doc| {
```
dav1do, line 72: why is this needed? shouldn't we already have a `DidDocument` built from ID? Is there a type mismatch or do we not actually need to build it twice (it compiled both ways for me.. so I'm not sure if there's something subtle/annoying going on)? I didn't run a `cargo update`.

runner/src/simulate.rs
```diff
@@ -617,6 +628,17 @@ impl ScenarioState {
                     Err(e) => (CommandResult::Failure(e), None),
                 }
             }
+            Scenario::CASBenchmark => {
+                let ids = self
```
dav1do, line 632: Is this just for debugging/understanding things and we'll enforce the limits later?

dav1do (approved):
A few comments, but I don't mind if you merge into the other branch and modify them all at once. You will have to sort out the conflicts (and possibly the clippy lint) before you can merge though.

runner/src/scenario/ceramic/mod.rs
samika98: I was getting conflicting types of DidDocument imports but these somehow resolved itself... can be removed I guess

runner/src/scenario/ceramic/anchor.rs
```diff
@@ -0,0 +1,174 @@
...
+        root_cid.clone(),
+    )
+    .await
+    .unwrap();
+    let cas_create_request_url = cas_service_url.clone() + "/api/v0/requests";
+    let goose_request = GooseRequest::builder()
+        .name("create_anchor_request")
+        .method(GooseMethod::Post)
+        .set_request_builder(
+            reqwest::Client::new()
```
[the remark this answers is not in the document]
samika98: Yes, makes sense. I was trying to figure out why we couldn't generate higher load in this scenario, this might be the reason why! Thanks!

PR #165, opened 2024-04-24
chore: disable goose logs by default

dav1do:
Don't write goose log files on workers unless requested. This will write goose logs to the manager stdout as well at the desired level. The simulation spec has a new field `logLevel: warn/info/debug/trace` and defaults to None (meaning off) if not set. This is a change, but these files haven't been generally useful afaik and it seems to consume a lot of RAM as they don't appear to be flushed until the end of the run. 

This requires redeploying the CRDs and the operator and runner images.

### DIFF
--- a/operator/src/simulation/spec.rs
+++ b/operator/src/simulation/spec.rs
@@ -34,6 +34,11 @@ pub struct SimulationSpec {
     /// Enable dev mode for the simulation. This will remove resource limits that are not
     /// explicitly set in the simulation spec.
     pub(crate) dev_mode: Option<bool>,
+    /// Log level to use. On the manager writes to stdout, and on the workers will write
+    /// files for the simulation. The files may not be flushed until the simulation is complete
+    /// which consumes RAM, and will disappear if there is no persistent volume when the pod exits.
+    /// Valid values: 'off', 'warn', 'info', 'debug', 'trace'. Defaults to 'off'.
+    pub(crate) log_level: Option<String>,
 }
 
 impl Simulation {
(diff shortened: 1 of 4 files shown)


runner/src/simulate.rs
```diff
@@ -83,6 +87,31 @@ pub struct Opts {
...
+}
+
+impl LogLevel {
+    fn is_enabled(&self) -> bool {
+        !matches!(self, LogLevel::Off)
+    }
+
+    fn as_goose_log_level(&self) -> u8 {
+        match self {
+            LogLevel::Off => panic!("LogLevel::Off should not be used as an integer (it is None)."),
```
3benbox, line 106: uhm, huh? This is mimicking goose functionality?

3benbox (approved):
lgtm

runner/src/simulate.rs
dav1do: 🙈 yeah, this was dumb. I thought I needed a value at first, so rather than use None I set it to off and then did this weird, "call this function to make sure it's not one variant and then do normal things. If you didn't follow this contract I will blow up in your face!" This was to deal with the fact that you can't set `GooseConfig.log_level = None`, but it turns out there's a much easier way.. which is use none lol. So now it does that.

operator/src/simulation/spec.rs
```diff
@@ -34,6 +34,11 @@ pub struct SimulationSpec {
     /// Enable dev mode for the simulation. This will remove resource limits that are not
     /// explicitly set in the simulation spec.
     pub(crate) dev_mode: Option<bool>,
+    /// Log level to use. On the manager writes to stdout, and on the workers will write
+    /// files for the simulation. The files may not be flushed until the simulation is complete
+    /// which consumes RAM, and will disappear if there is no persistent volume when the pod exits.
```
samika98, line 39: More of an exploratory question: Can we define constraints on memory while running a simulation? In the resource definitions of the managers/workers pods I did not notice and resource constraint fields. How do we define what resources to give to a manger/worker during a simulation run

PR #167, opened 2024-05-01
feat: allow overriding cas api envars

smrz2001:
I moved the env overriding logic to a utility function, and included the sorting in there as well. This caused a lot of the test files to churn, but hopefully this will make them more consistent going forward.

### DIFF
--- a/operator/src/network/cas.rs
+++ b/operator/src/network/cas.rs
@@ -30,8 +30,9 @@ use crate::{
         ipfs::{IpfsConfig, IpfsInfo, IPFS_DATA_PV_CLAIM},
         node_affinity::NodeAffinityConfig,
         resource_limits::ResourceLimitsConfig,
-        CasSpec,
+        CasApiSpec, CasSpec,
     },
+    utils::override_env_vars,
 };
 
 const CAS_IPFS_INFO_SUFFIX: &str = "cas";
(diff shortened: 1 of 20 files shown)


operator/src/utils/mod.rs
```diff
@@ -318,3 +318,21 @@ pub fn generate_random_secret(
     rng.fill_bytes(&mut secret_bytes);
     hex::encode(secret_bytes)
 }
+
+/// Apply override env vars to an existing env var list
+pub fn override_env_vars(env: &mut Vec<EnvVar>, overrides: Option<BTreeMap<String, String>>) {
```
dav1do, line 323: think this could be `&Option<BTreeMap<String, String>>` since you take a reference to iterate and it would remove the clone.. it isn't important for perf, but I like to make the api match the requirement (i.e make the caller clone if I needed owned, or take a ref if we don't.. but taking a ref and then cloning is usually a bad sign).

operator/src/network/cas.rs
```diff
@@ -44,6 +45,18 @@ pub struct CasConfig {
     pub ganache_resource_limits: ResourceLimitsConfig,
     pub postgres_resource_limits: ResourceLimitsConfig,
     pub localstack_resource_limits: ResourceLimitsConfig,
+    pub cas_api: CasApiConfig,
```
dav1do, line 48: maybe update the guide with an example of this? I think the section explaining casResourceLimits could maybe have it? I think it's like this?

```yaml
# network configuration
---
kind: Network
metadata:
  name: small
spec:
  replicas: 2
  cas:
    image: ceramicnetwork/ceramic-anchor-service:latest
    casResourceLimits:
      cpu: "250m"
      memory: "1Gi"
    casApi:
      env: ...
 ```

dav1do (approved):
Minor comments :shipit:

PR #168, opened 2024-05-02
fix: always sort env vars after override

smrz2001:
Moved env var sorting outside the if-check 😅


dav1do (approved):
😅 oops, I didn't notice either

PR #170, opened 2024-05-06
feat/new model_reuse_type to enable sync by peers and creation by leader

samika98:
Added a Model Reuse Type to help us configure Goose Scenario. This Reuse type lets the global leader (worker with id 0) create models and all the workers (users) subscribe to that model.

### DIFF
--- a/runner/src/scenario/ceramic/model_instance.rs
+++ b/runner/src/scenario/ceramic/model_instance.rs
@@ -154,8 +154,15 @@ impl CeramicModelInstanceTestUser {
         debug!(params=?config.params, "setting up scenario");
         let (small_model_id, large_model_id) = match config.params.model_reuse {
             super::ReuseType::PerUser => {
-                Self::generate_list_models(user, &config.admin_cli, &config.redis_cli, true, None)
-                    .await?
+                Self::generate_list_models(
+                    user,
+                    &config.admin_cli,
+                    &config.redis_cli,
+                    true,
+                    None,
+                    false,
+                )
+                .await?
             }
             super::ReuseType::Shared => {
                 let (small, large) = Self::generate_list_models(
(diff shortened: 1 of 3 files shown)


runner/src/scenario/ceramic/model_instance.rs
```diff
@@ -251,6 +274,20 @@ impl CeramicModelInstanceTestUser {
...
+            super::ReuseType::LeadWorkerSubscriber => {
+                // For model instance reuse type make it work the same way as shared
+                Self::generate_mids(
+                    user,
+                    &config.user_cli,
+                    &config.redis_cli,
+                    &small_model_id,
+                    &large_model_id,
+                    config.params.number_of_documents,
+                    global_leader,
```
smrz2001, line 286: This should be `true` so that each worker creates MIDs instead of just the global leader.
```suggestion
                    true,
```

runner/src/scenario/ceramic/model_instance.rs
```diff
@@ -180,9 +188,24 @@ impl CeramicModelInstanceTestUser {
...
+                    true,
+                )
+                .await?;
+                Self::ensure_model_exists(user, &config.user_cli, &small).await?;
+                Self::ensure_model_exists(user, &config.user_cli, &large).await?;
+                (small, large)
+            }
         };
 
         if lead_user {
```
smrz2001, line 211: Because of this `if`-block, one user per worker will call the model subscription API, which means each nod will already be subscribed to the models. This means that we shouldn't need to pass the `should_subscribe` flag as you're doing below.

runner/src/scenario/ceramic/model_instance.rs
smrz2001: Basically, we need the `ReuseType::Shared` code when creating models, but `ReuseType::PerUser` code when creating MIDs.

runner/src/scenario/ceramic/model_instance.rs
```diff
@@ -293,6 +330,7 @@ impl CeramicModelInstanceTestUser {
         redis_cli: &redis::Client,
         should_create: bool,
         redis_postfix: Option<String>,
+        should_subscribe: bool,
```
smrz2001, line 333: This might not be necessary - see my comment above. One user on each worker is already making sure that the corresponding node is subscribed to the models.

PR #171, opened 2024-05-10
chore: use "real" car files for events in scenarios

dav1do:
This fixes the errors running the recon scenario on new versions of ceramic one by creating car files that look like real events so they pass the basic validation. This also puts the actual root CID of the block in c1, so we no longer see the "block not found locally" errors that we were getting in js-ceramic. 

I want to clean up the duplicate code and hacks around cid/multihash/ipld versions, and it should be easy once the [http client](https://github.com/3box/ceramic-http-client-rs/pull/17) changes land, but for now getting this working is more important.


samika98 (approved):
LGTM!

PR #176, opened 2024-05-14
feat: enable ceramic nodejs inspection

smrz2001:


operator/src/network/ceramic.rs
```diff
@@ -463,6 +471,28 @@ pub fn stateful_set_spec(ns: &str, bundle: &CeramicBundle<'_>) -> StatefulSetSpe
...
     volumes.append(&mut bundle.config.ipfs.volumes(&bundle.info));
 
+    let mut ceramic_ports = vec![
+        ContainerPort {
+            container_port: CERAMIC_SERVICE_API_PORT,
+            name: Some("api".to_owned()),
+            ..Default::default()
+        },
+        ContainerPort {
+            container_port: 9464,
```
dav1do, line 481: Any reason to make these ports (9229 and 9464) constants like `CERAMIC_SERVICE_API_PORT`? If this is the only place they're used, probably not necessary just curious.

dav1do (approved):
🚀

operator/src/network/ceramic.rs
smrz2001: Yeah, good idea. 9464 is the default metrics port and used in a bunch of places. And 9229 is the standard node inspection port.

I'll make consts for them.

PR #177, opened 2024-05-16
fix: update default resources for rust ipfs

nathanielc:
As we know having less than 1 cpu can cause the odd deadlock in the C1 process. This makes the default a sane default.

### DIFF
--- a/operator/src/network/ipfs.rs
+++ b/operator/src/network/ipfs.rs
@@ -119,8 +119,8 @@ impl Default for RustIpfsConfig {
             image: "public.ecr.aws/r5b3e0r5/3box/ceramic-one:latest".to_owned(),
             image_pull_policy: "Always".to_owned(),
             resource_limits: ResourceLimitsConfig {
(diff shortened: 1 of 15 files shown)


operator/src/network/ipfs.rs
```diff
@@ -119,8 +119,8 @@ impl Default for RustIpfsConfig {
             image: "public.ecr.aws/r5b3e0r5/3box/ceramic-one:latest".to_owned(),
             image_pull_policy: "Always".to_owned(),
             resource_limits: ResourceLimitsConfig {
-                cpu: Some(Quantity("250m".to_owned())),
-                memory: Some(Quantity("512Mi".to_owned())),
+                cpu: Some(Quantity("1".to_owned())),
+                memory: Some(Quantity("1Gi".to_owned())),
```
nathanielc, line 123: This is the only real, change. The rest are test updates.

smrz2001 (approved):
🚀

PR #178, opened 2024-05-21
feat: support configuring persistent storage for containers

dav1do:
Add persistent storage config object that allows modifying the persistent volume storage size and class. I updated a handful of tests to set this, so we see the 10 -> override value, demonstrating that the default is still 10 (and shown in tests that haven't been updated). If preferable, I can write dedicated tests for this new setting.

For IPFS, we allowed specifying `storageClass` before. We still support this field, but will prefer the `storage.class` field if it is set. If not, we will use `storageClass` (there are existing tests that demonstrate this).

### DIFF
--- a/keramik/src/advanced_configuration.md
+++ b/keramik/src/advanced_configuration.md
@@ -70,6 +70,45 @@ and adding the following environment variables to the `spec/template/spec/contai
 
 # Image Resources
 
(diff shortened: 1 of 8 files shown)


3benbox (approved):
LGTM!

PR #180, opened 2024-05-21
fix: use different aws region for cas than us-east-1

smrz2001:
The CAS Scheduler specifies a location constraint when creating a S3 bucket. `us-east-1` is not a valid option when specifying the region.


JulissaDantes (approved):
Thank you!

PR #182, opened 2024-06-06
chore: use new ceramic one metric for validation

dav1do:
Update metric to match renames in [rust ceramic](https://github.com/ceramicnetwork/rust-ceramic/pull/373). We fall back to old one if new doesn't exist so it should pass in CI/tests for either version of c1.


smrz2001 (approved):
LGTM!

samika98 (approved):
lgtm

3benbox (approved):
lgtm

PR #183, opened 2024-06-11
chore: update js-ceramic healthcheck timing

3benbox:
- lower initial_delay_seconds so that healthchecks are run faster and pods are marked up earlier
- lower period_seconds so that there are less healthchecks (already enough load on the system!)
- increase timeout_seconds as the js-ceramic health check makes its own external calls to "ipfs" which where longer than this setting


dav1do (approved):
This seems aok and I like reducing the `initial_delay_seconds` but find it hard to believe the timeouts and period were too low (it makes me sad).

We could consider using different values for startup and liveness [like this](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes) if we're still seeing odd behavior, or possibly adjust the `failureThreshold`.

PR #184, opened 2024-06-12
feat: aes 84 validate correctness in keramik ceramic anchoring benchmark

samika98:
Features split in this PR : 

- [x] This PR : Add validation to check if all streams are anchored after a configurable run-time. This run-time should be passed in the scenario spec yaml while running the scenario ceramic-anchoring-benchmark
- [x] This PR: Change cas-benchmark scenario to pass cas-network in the scenario yaml. Use this specified field to decide the cas network to test against. By default use dev. This configuration is optional
- [x] This PR : Pass in cas-controller from spec for cas-benchmark, cas-controller will be a did:key
- [x] Verify sync in ceramic-anchoring-benchmark - PR : https://github.com/3box/keramik/pull/185
- [ ] Convert this into a nightly perf test - NEXT PR


runner/src/scenario/ceramic/model_instance.rs
```diff
@@ -73,6 +73,7 @@ pub struct EnvBasedConfig {
 
 #[derive(Clone, Debug)]
 pub struct GooseUserInfo {
+    #[allow(dead_code)]
```
samika98, line 76: rust updated on CI and figured out that these variables were not being read. @dav1do is that intentional are these present for future use?

runner/src/scenario/ceramic/mod.rs
```diff
@@ -21,6 +21,7 @@ pub type CeramicClient = CeramicHttpClient<JwkSigner>;
 #[derive(Clone)]
 pub struct Credentials {
     pub signer: JwkSigner,
+    #[allow(dead_code)]
```
samika98, line 24: Same here @dav1do

runner/src/utils.rs
```diff
@@ -52,3 +54,57 @@ pub async fn parse_peers_info(path: impl AsRef<Path>) -> Result<Vec<Peer>> {
...
+    population_size: usize,
+    confidence_level: f64,
+    margin_of_error: f64,
+) -> usize {
+    let mut z = 2.576; // Z-score for 99% confidence interval
+    if confidence_level == 0.95 {
+        z = 1.96;
+    } else if confidence_level == 0.99 {
+        z = 2.576;
+    } else {
```
dav1do, line 77: I tend to prefer using if/let assignment to mutability but same same

```rust
    let z = if confidence_level == 0.95 {
        1.96
    } else {
        error!(
            "Invalid confidence level: {}, defaulting to 0.99",
            confidence_level
        );
        2.576 // Z-score for 99% confidence interval
    }
```

runner/src/simulate.rs
```diff
@@ -700,21 +730,33 @@ impl ScenarioState {
...
-        // Pick a peer at random
+        info!("Waiting for {} seconds", wait_duration.as_secs());
+        // Pick the first peer as the validator. It should have all the streams synced to it
         let peer = self.peers.first().unwrap();
 
         let ids = self.get_set_from_redis(ANCHOR_REQUEST_MIDS_KEY).await?;
         info!("Number of MIDs: {}", ids.len());
 
+        // Calculate the sample size
+        let sample_size = calculate_sample_size(ids.len(), 0.99, 0.02).await;
```
dav1do, line 750: nice, this is fancy :) I would have probably just done `iter_by(1000)` on the first pass 😅

runner/src/scenario/ceramic/model_instance.rs
dav1do: That is really weird.. both of these are used by scenarios currently... maybe we can pair quickly to see what's up? I'm not seeing any clippy lints locally so I'm a bit confused. This is the only thing I'd like to sort out before merging, the other comments are just observations.

dav1do (commented):
Looking good! would like to chat about the `dead_code` but otherwise just comments for thought 😄

runner/src/simulate.rs
```diff
@@ -88,8 +88,19 @@ pub struct Opts {
     /// left to the scenario (requests per second, total requests, rps/node etc).
     #[arg(long, env = "SIMULATE_TARGET_REQUESTS")]
     target_request_rate: Option<usize>,
-    // #[arg(long, env = "SIMULATE_ANCHOR_WAIT_TIME")]
-    // anchor_wait_time: Option<u64>,
+
+    // Wait time after which we wosh to check the anchor correctness.
```
dav1do, line 92: typo wosh -> wish. Do you mind adding a comment about it's purpose (e.g. the scenario will not finish until this met as we wait for all streams to be anchored or whatever it is specifically)

runner/src/simulate.rs
```diff
@@ -700,21 +730,33 @@ impl ScenarioState {
...
+
+        // Select a random sample set of IDs
+        let sample_ids = select_sample_set_ids(&ids, sample_size).await;
+        info!("Sampled IDs count: {}", sample_ids.len());
+
         // Make an API call to get the status of request from the chosen peer
-        for stream_id in ids.clone() {
-            info!("Fetching anchor status for streamID {}", stream_id);
+        for stream_id in sample_ids.iter() {
+            // Only fetch anchor status for every thousandth stream ID
```
dav1do, line 759: I don't think this comment is accurate with the stats stuff in the mix

samika98: It is not, will remove it

dav1do (approved):

PR #185, opened 2024-06-17
feat: aes-135-define-a-network-where-we-only-write-to-one-node-and-read

samika98:
Check the parent for description : https://github.com/3box/keramik/pull/184

### DIFF
--- a/runner/Cargo.toml
+++ b/runner/Cargo.toml
@@ -30,6 +30,7 @@ serde_ipld_dagcbor = "0.6"
 serde_ipld_dagjson = "0.2"
 schemars.workspace = true
 serde_json.workspace = true
+once_cell = "1.19.0"
 tokio.workspace = true
 tracing.workspace = true
 tracing-subscriber.workspace = true
--- a/runner/src/scenario/ceramic/new_streams.rs
+++ b/runner/src/scenario/ceramic/new_streams.rs
@@ -7,7 +7,7 @@ use tracing::info;
 
 use crate::scenario::{
     ceramic::{model_instance::CountResponse, models, RandomModelInstance},
-    get_redis_client,
+    get_redis_client, is_goose_lead_worker,
 };
 
 use super::{
@@ -144,7 +144,7 @@ pub async fn small_large_scenario(
     let instantiate_small_model = Transaction::new(Arc::new(move |user| {
         let small_model_conn_clone = small_model_conn.clone();
         Box::pin(async move {
-            instantiate_small_model(user, params.store_mids, small_model_conn_clone).await
+            instantiate_small_model(user, params.store_mids, small_model_conn_clone, false).await
         })
     }))
     .set_name("instantiate_small_model");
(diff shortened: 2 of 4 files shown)


runner/src/scenario/ceramic/new_streams.rs
```diff
@@ -163,6 +163,36 @@ pub async fn small_large_scenario(
         .register_transaction(instantiate_large_model))
 }
 
+pub async fn small_scenario(params: CeramicScenarioParameters) -> Result<Scenario, GooseError> {
+    let redis_cli = get_redis_client().await.unwrap();
+    let multiplexed_conn = redis_cli.get_multiplexed_tokio_connection().await.unwrap();
+    let shared_conn = Arc::new(Mutex::new(multiplexed_conn));
```
dav1do, line 169: i'm a little worried about putting an active connection in here for a long test.. i'm not sure how stable the redis client is and feel like I've seen weird dead connection issues but maybe i'm making that up. I'd probably prefer to just pull the redis client out of the user data and make a new connection unless the perf overhead is too much. we can try this first

runner/src/scenario/ceramic/new_streams.rs
```diff
@@ -207,20 +237,24 @@ async fn instantiate_small_model(
...
+            user_data.user_cli(),
+            &user_data.small_model_id,
+            "instantiate_small_model_instance",
+            &models::SmallModel::random(),
+        )
+        .await?;
+        if store_in_redis {
+            let mut conn: tokio::sync::MutexGuard<'_, MultiplexedConnection> = conn.lock().await;
+            let stream_id_string = response.to_string();
+            let _: () = conn.sadd("anchor_mids", stream_id_string).await.unwrap();
```
dav1do, line 256: continuing my comment from above, if the connection is dead, we could grab `user_data.redis_cli()` and make/store a new one

runner/src/simulate.rs
```diff
@@ -695,6 +697,39 @@ impl ScenarioState {
         Ok(metrics.state.anchor_status)
     }
 
+    pub async fn model_sync_status(
+        &self,
+        peer: &Peer,
+        stream_id: String,
+    ) -> Result<bool, anyhow::Error> {
+        let client = reqwest::Client::new();
```
dav1do, line 705: creating a new client is heavy (it has an internal threadpool). we should get one of out of the user_data object/a local static/or pass it in as a parameter.

See https://docs.rs/reqwest/latest/reqwest/struct.Client.html specifically `The Client holds a connection pool internally, so it is advised that you create one and reuse it.`

runner/src/scenario/ceramic/new_streams.rs
```diff
@@ -207,20 +237,24 @@ async fn instantiate_small_model(
     user: &mut GooseUser,
     store_in_redis: bool,
     conn: Arc<tokio::sync::Mutex<MultiplexedConnection>>,
+    only_once_per_network: bool,
```
dav1do, line 240: `only_once_per_network` confused me at first, I was thinking it was once globally, not one writer... maybe we can call it one_sided_test or one_writer_per_network or something? once makes me think literally once and that doesn't seem to be the intention.

dav1do (commented):
nice! have a few questions/comments but nothing major 🚀

runner/src/simulate.rs
samika98: I do not see how we can use the GooseUser outside a GooseTransaction. We can create a static client and use that. For now I can create it in simulate.rs

runner/Cargo.toml
```diff
@@ -30,6 +30,7 @@ serde_ipld_dagcbor = "0.6"
 serde_ipld_dagjson = "0.2"
 schemars.workspace = true
 serde_json.workspace = true
+once_cell = "1.19.0"
```
dav1do, line 33: nice, just an fyi you can use a [OnceLock](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) from the std library now, but I prefer the API on once_cell tbh

dav1do (approved):

PR #187, opened 2024-06-18
chore: update docs to include simulation specs

samika98:
Updates the docs to include scenarios added and inclides instructions on how to run them
Scenarios : 
    1) ceramic-anchoring-benchmark : This is the e2e test which creates models and passes when sync on the network is working and all streams have been anchored. We should set the CAS Api url in the network with proper dids to test against the CAS. 
```

  casApiUrl: https://cas-dev-direct.3boxlabs.com
  networkType: dev-unstable
  privateKeySecret: ceramic-v4-dev
```

 New Simulation vars for ceramic-anchoring-benchmark : 

        ` anchorWaitTime : Wait time in seconds for how long we want to wait after streams have been created to check when they have been anchored. This should be a high number like 30-40 minutes.`

  2) cas-benchmark : This tests the write CAS load directly by passing c1 nodes. New Simulation vars for cas-benchmark : 

```
          casNetwork : Network-url. Similar to CAS api url. We can use dev-direct if we wish to bypass auth here.

          casController : Did key we want to provide in the network. Format : "did:key:<secret>"
```

### DIFF
--- a/keramik/src/simulation.md
+++ b/keramik/src/simulation.md
@@ -8,6 +8,8 @@ To run a simulation, first define a simulation. Available simulation types are
 - `ceramic-new-streams` - A simulation that only creates new streams
 - `ceramic-model-reuse` - A simulation that reuses the same model and queries instances across workers
 - `recon-event-sync` - A simulation that creates events for Recon to sync at a fixed rate (~300/s by default). Designed for a 2 node network but should work on any.
+- `cas-benchmark` - A simulation that benchmarks the CAS network. 
+- `cas-anchoring-benchmark` - A simulation that benchmarks the Ceramic with anchoring enabled.
 
 Using one of these scenarios, we can then define the configuration for that scenario:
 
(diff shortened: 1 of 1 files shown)


keramik/src/simulation.md
```diff
@@ -124,3 +126,91 @@ spec:
...
+          env:
+            CERAMIC_ONE_RECON: "true"
+  casApiUrl: https://cas-dev-direct.3boxlabs.com
+  networkType: dev-unstable
+  privateKeySecret: ceramic-v4-dev
+  ethRpcUrl: ""
+```
+
+```shell
+kubectl apply -f custom-ipfs.yaml
```
3benbox, line 155:
```suggestion
kubectl apply -f custom-cas-api.yaml
```

3benbox (approved):
Thank you for the docs!

1 suggestion

stephhuynh18 (approved):

PR #189, opened 2024-06-24
chore: update ceramic http client and fix recon sync test

dav1do:
Updates the `ceramic-http-client-rs` dependency and uses the new event format from rust ceramic. This allows the `recon-event-sync` scenario to work without failing due "unexpected fields" when the API roundtrips the payload to make sure it is as expected.

### DIFF
--- a/runner/src/scenario/ceramic/model_instance.rs
+++ b/runner/src/scenario/ceramic/model_instance.rs
@@ -688,7 +688,6 @@ impl ModelInstanceRequests {
         let name = format!("create_index_model_{}", tx_name);
         let req = cli
             .create_index_model_request(model_id, &resp.code)
-            .await
             .unwrap();
         let mut goose = user
             .request(
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -703,11 +703,10 @@ dependencies = [
 
 [[package]]
 name = "ceramic-core"
-version = "0.18.0"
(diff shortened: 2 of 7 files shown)


smrz2001 (approved):
🚀

runner/src/scenario/ceramic/model_instance.rs
```diff
@@ -688,7 +688,6 @@ impl ModelInstanceRequests {
         let name = format!("create_index_model_{}", tx_name);
         let req = cli
             .create_index_model_request(model_id, &resp.code)
-            .await
```
samika98, line 691: Why did we remove the await here? Isn't this method async?

samika98 (commented):
Everything else looks good just a question about removing an await

runner/src/scenario/ceramic/model_instance.rs
dav1do: not anymore. the signer used by ceramic_core changed in [this PR](https://github.com/ceramicnetwork/rust-ceramic/pull/357/files#diff-ea616cf3110776cfec65245b2c2def84dd0f227005583ca991b8755838fa54c3) and isn't async anymore.. so the http client was updated and finally this one

dav1do: you should get a warning if you don't ever await a future, which should cause our CI to fail since we deny warnings

PR #191, opened 2024-06-25
feat: make anchoring fast by default

nathanielc:
When keramik is used to deploy CAS it is only in the context of a hermetic testing env. Therefore we can assume that anchoring should be as fast as possible in order to all for faster tests etc.

Makes two changes:

1. Reduces worker delay to 1s instead of 1m after completing a batch.
2. Configures ganache to instantly mine any transaction.


smrz2001 (approved):
🚀

PR #193, opened 2024-06-25
fix: use smallest time for block time

nathanielc:
Using the instamine option of ganache confuses the CAS worker and it waits forever to see the transaction has been mined. This change uses a block time of 1 second, however in practice this appears to mean a block every 15 seconds. Faster than once a minute but still not very fast.


smrz2001 (approved):
🚀

PR #197, opened 2024-07-03
fix: set kubo API port to 5101

dav1do:
We need to override the kubo API port at start up to match the port we're opening on the container.


smrz2001 (approved):
🚀

PR #198, opened 2024-07-03
feat: otel prometheus update

3benbox:
- update opentelemetry image for prometheusremotewrite exporter
- update prometheus image for remote write ability
- deploy prometheus service for opentelemetry remote write
- move configmap data to yaml file for easier operations
- remove extra opentelemtry exporter for prometheus/simulations (didn't seem used?)

Prometheus doesn't scrape anything now, it is pushed from opentelemetry.
This should reduce duplicate metrics and provide a single collector to forward to different exporters as required.

### DIFF
--- a/operator/src/monitoring/opentelemetry.rs
+++ b/operator/src/monitoring/opentelemetry.rs
@@ -9,7 +9,7 @@ use k8s_openapi::{
             PodSpec, PodTemplateSpec, ResourceRequirements, ServicePort, ServiceSpec, Volume,
             VolumeMount,
         },
-        rbac::v1::{PolicyRule, RoleRef, Subject, Role, RoleBinding},
+        rbac::v1::{PolicyRule, Role, RoleBinding, RoleRef, Subject},
     },
     apimachinery::pkg::{
         api::resource::Quantity,
@@ -27,7 +27,7 @@ use crate::{
         resource_limits::ResourceLimitsConfig,
     },
     utils::{
-        apply_account, apply_namespaced_role, apply_namespaced_role_binding, apply_config_map,
+        apply_account, apply_config_map, apply_namespaced_role, apply_namespaced_role_binding,
         apply_service, apply_stateful_set, Clock, Context,
     },
 };
@@ -59,7 +59,8 @@ pub async fn apply(
         orefs.to_vec(),
         OTEL_ROLE_BINDING,
         role_binding(ns),
-    ).await?;
+    )
+    .await?;
     apply_config_map(
         cx.clone(),
         ns,
@@ -256,7 +257,6 @@ fn stateful_set_spec(config: &OtelConfig) -> StatefulSetSpec {
         ..Default::default()
     }
 }
-
 
 fn config_map_data() -> BTreeMap<String, String> {
     let config_str = include_str!("./otel-config.yaml"); // Adjust the path as necessary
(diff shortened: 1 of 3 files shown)


operator/src/monitoring/prom-config.yaml
```diff
@@ -0,0 +1,4 @@
+global:
+  scrape_interval: 10s
+  scrape_timeout: 5s
```
samika98, line 3: These seem a bit aggressive to me. Do we need near real-time metrics for our use cases? Can we scrape every 30 seconds to a minute?

nathanielc (approved):

operator/src/monitoring/otel-config.yaml
```diff
@@ -0,0 +1,82 @@
+---
+receivers:
+  # Push based metrics
+  otlp:
```
samika98, line 4: Which services use the push-based metrics vs which services set up the metrics that we are scraping using prometheus? Can these metrics be colliding? Can we add a comment or doc in terms of how we should name/label metrics to avoid such collisions

operator/src/monitoring/otel-config.yaml
```diff
@@ -0,0 +1,82 @@
...
+            action: replace
+            target_label: kubernetes_namespace
+          - source_labels: [__meta_kubernetes_pod_name]
+            action: replace
+            target_label: kubernetes_pod
+          - source_labels: [__meta_kubernetes_pod_container_name]
+            action: replace
+            target_label: kubernetes_container
+
+processors:
```
samika98, line 46: Can we filter out duplicates while processing?

samika98 (changes requested):
Looks good to me! Have a questions/concerns about having both push and pull based metrics in the system

operator/src/monitoring/prom-config.yaml
3benbox: I just pulled these setting along from the existing config.
Want to drop them down to the defaults?

```
# How frequently to scrape targets by default.
  [ scrape_interval: [<duration>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#duration) | default = 1m ]

  # How long until a scrape request times out.
  [ scrape_timeout: [<duration>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#duration) | default = 10s ]
  ```

operator/src/monitoring/otel-config.yaml
3benbox: Scraping will be performed on any pods in the namespace that have a port name "metrics"
https://github.com/3box/keramik/pull/198/files#diff-343efbc6f501d7065cfa0c42c6d66ea7b49ab1783578c7b695b6acf4c1728400R11-R45

Push based metrics would be anyone sending metric to the collector.
In keramik, I believe, and would like, that to be just the simulation manager and workers.
Since they're ephemeral jobs, they should push their metrics before the go away/stop.

3benbox: I think if we keep this separation, services pull, simulations push, we should avoid duplications.

operator/src/monitoring/otel-config.yaml
3benbox: We can if necessary, see this issue about the situation.

https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/14900

samika98 (approved):

PR #199, opened 2024-07-09
fix: pass cas api cpus explicitly to the process

smrz2001:
The code for CAS API clustering (launching as many NodeJS processes as their are CPUs available) is not working in Keramik. This is causing the host's CPUs to be used, which launches 32 processes and causes the container to OOM.

This PR converts the CAS resource limits to an equivalent number of CPUs with a minimum of 1, and passes this number in to the container explicitly.


3benbox (approved):
LGTM!

PR #200, opened 2024-07-09
fix: use cas develop image by default

smrz2001:
Need CAS API changes currently in develop, but also probably a good idea to use the develop image in general.


samika98 (approved):
LGTM!

PR #202, opened 2024-07-12
fix: use composedb develop image by default

smrz2001:
This aligns all the default versions of the software we set up in Keramik: `ceramic-one:latest` / `js-ceramic:develop` / `composedb:develop` / `ceramic-anchor-service:develop`.

Our tests will usually specify explicit versions of the software being tested.

Thanks for bringing this up, @samika98.

### DIFF
--- a/local-scripts/load.py
+++ b/local-scripts/load.py
@@ -2,11 +2,11 @@ import subprocess
 
 images = [
         'amazon/aws-cli',
-        'ceramicnetwork/ceramic-anchor-service:latest',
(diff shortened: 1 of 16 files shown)


samika98 (approved):
LGTM! 
nit -> The load.py in the local-scripts dir still has some old images. It would be nice if we can set the images there to the defaults as well. https://github.com/3box/keramik/blob/b2429b3272c0c1c29a81bff67b8417fcfd2869c7/local-scripts/load.py#L3

PR #203, opened 2024-07-12
feat: add generator code

samika98:
This PR introduces the initial logic for the runner. It includes the following key features:
- Adds a new command to the runner.
- Implements functionality to generate load against Ceramic instances.

This PR does not introduce any functional changes. It primarily adds interfaces and structures that will be used in future work. Once the operator changes are made, we will be able to call this command effectively.

This PR is part of a larger effort and contains several TODOs that will be addressed in subsequent PRs. 
For a quick review, the focus is on the runner logic and the new command implementation, rust coding practices/improvements.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ CARGO = CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse RUSTFLAGS="--cfg tokio_unstab
 all: build check-fmt check-clippy test
 
 .PHONY: test
-test:
+test: update
 	# Test with default features
 	${CARGO} test --locked
 	# Test with all features
@@ -13,15 +13,15 @@ test:
 	${CARGO} test --locked --no-default-features
 
 .PHONY: check-fmt
-check-fmt:
+check-fmt: update
 	cargo fmt --all -- --check
 
 .PHONY: update
 update:
 	${CARGO} update
 
-.PHONY: check-clippy
-check-clippy:
+.PHONY: check-clippy 
+check-clippy: update
 	# Check with default features
 	${CARGO} clippy --workspace
 	# Check with all features
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 CARGO = CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse RUSTFLAGS="--cfg tokio_unstable -D warnings" cargo
 
 .PHONY: all
-all: update build check-fmt check-clippy test
+all: build check-fmt check-clippy test
 
 .PHONY: test
 test:
@@ -16,6 +16,10 @@ test:
 check-fmt:
 	cargo fmt --all -- --check
 
+.PHONY: update
+update:
+	${CARGO} update
+
 .PHONY: check-clippy
 check-clippy:
 	# Check with default features
@@ -24,7 +28,7 @@ check-clippy:
 	${CARGO} clippy --workspace --all-features
 
 .PHONY: build
-build: runner operator
+build: update runner operator
 
 .PHONY: runner
 runner:
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ test: update
 	${CARGO} test --locked --no-default-features
 
 .PHONY: check-fmt
-check-fmt: update
+check-fmt:
 	cargo fmt --all -- --check
 
 .PHONY: update
@@ -21,14 +21,14 @@ update:
 	${CARGO} update
 
 .PHONY: check-clippy 
-check-clippy: update
+check-clippy:
 	# Check with default features
 	${CARGO} clippy --workspace
 	# Check with all features
 	${CARGO} clippy --workspace --all-features
 
 .PHONY: build
-build: update runner operator
+build: runner operator
 
 .PHONY: runner
 runner:
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 CARGO = CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse RUSTFLAGS="--cfg tokio_unstable -D warnings" cargo
 
 .PHONY: all
-all: build check-fmt check-clippy test
+all: update build check-fmt check-clippy test
 
 .PHONY: test
 test:
--- a/runner/src/load_generator/gen.rs
+++ b/runner/src/load_generator/gen.rs
@@ -1,31 +1,31 @@
 use std::collections::HashMap;
 use std::path::PathBuf;
 
+use crate::load_generator::utils::generator_utils::CeramicConfig;
+use crate::load_generator::utils::generator_utils::CeramicDidType;
+use crate::load_generator::utils::generator_utils::CeramicScenarioParameters;
+use crate::load_generator::utils::generator_utils::StableLoadUser;
+use crate::utils::parse_peers_info;
+use crate::CommandResult;
 use anyhow::Result;
 use ceramic_core::StreamId;
 use clap::Args;
 use keramik_common::peer_info::Peer;
-use tokio::time::{Duration, Instant};use crate::utils::parse_peers_info;
-use crate::CommandResult;
-use crate::load_generator::utils::generator_utils::StableLoadUser;
-use crate::load_generator::utils::generator_utils::CeramicConfig;
-use crate::load_generator::utils::generator_utils::CeramicDidType;
-use crate::load_generator::utils::generator_utils::CeramicScenarioParameters;
+use tokio::time::{Duration, Instant};
 
-// TODO : Use this to envoke a particular scenario, currently we only have one 
+// TODO : Use this to envoke a particular scenario, currently we only have one
 // so this is unused
+#[allow(dead_code)]
 pub enum WeekLongSimulationScenarios {
     CreateModelInstancesSynced,
 }
 
 /// Options to Simulate command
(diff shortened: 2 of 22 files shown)


runner/src/load_generator/gen.rs
```diff
@@ -0,0 +1,233 @@
...
+                    Ok(_) => {
+                        count += 1;
+                    }
+                    Err(err) => {
+                        *error_map.entry(err).or_insert(0) += 1;
+                    }
+                }
+            }
+        }
+        if start_time.elapsed() > duration {
```
dav1do, line 181: it might be nice to add at least a few seconds to this so you give the tasks a chance to exit gracefully as they're waiting for the exact same time. I'd suggest at least 5 since that's the timeout above

runner/src/load_generator/utils/ceramic_models_utils.rs
```diff
@@ -0,0 +1,171 @@
...
+            .unwrap();
+        let req = self
+            .ceramic_client
+            .create_index_model_request(model_id, &admin_code)
+            .unwrap();
+        let resp = self.http_client.post(url).json(&req).send().await?;
+        if resp.status().is_success() {
+            Ok(())
+        } else {
+            Err(anyhow::anyhow!("Failed to index model"))
```
dav1do, line 48: might be nice to include the model ID in this message

runner/src/load_generator/utils/ceramic_models_utils.rs
```diff
@@ -0,0 +1,171 @@
...
+
+    /**
+     * Generate a random model
+     *
+     * @return The stream id of the created model
+     */
+    pub async fn generate_random_model(&self) -> Result<StreamId, anyhow::Error> {
+        let small_model =
+            ModelDefinition::new::<SmallModel>("load_test_small_model", ModelAccountRelation::List)
+                .unwrap();
```
dav1do, line 60: can we use `?` or `map_err` throughout this file to remove the `unwrap()` calls? nicer to log an error than crash the generator

runner/src/load_generator/utils/generator_utils.rs
```diff
@@ -0,0 +1,105 @@
...
+use crate::scenario::ceramic::Credentials;
+use anyhow::Result;
+use ceramic_http_client::CeramicHttpClient;
+use reqwest::Client;
+use std::time::Duration;
+
+use super::ceramic_models_utils::CeramicModelUser;
+
+pub static HTTP_TIMEOUT: Duration = Duration::from_secs(5);
+pub static HTTP_POOL_MAX_IDLE_PER_HOST: usize = 300;
```
dav1do, line 11: these could be `const` but it doesn't matter in practice. `const` is inserted into the calling location at compile time, so it's more performant, `static` is one shared memory location. you can't use const for shared data as it won't do what you want, and not all functions are supported but this one should be fine

runner/src/load_generator/utils/mod.rs
```diff
@@ -0,0 +1,2 @@
+pub mod ceramic_models_utils;
```
dav1do, line 1: it might be nicer to export the types you want and just refer to `load_generator::utils::CeramicConfig` etc and remove the extra module in the path

rather than having a file called `generator_utils.rs` in a module called generator_utils, I'd probably end up putting it in mod.rs. that said, I generally prefer the lib/mod file to just be exports with all the code living in its own file, so I'd suggest at least erasing the name by exporting types directly. basically, you could change

```rust
pub mod ceramic_models_utils;
pub mod generator_utils;
```
to 
```rust
mod ceramic_models_utils;
mod generator_utils;

// pub use ::* isn't the best practice, but makes sense in a utils crate
pub use ceramic_models_utils::*;
pub use generator_utils::*;
```
and then in other places you'd just refer to `load_generator::utils::X`.. if models/generator stuff are distinct enough, they might merit their own modules rather than being deep inside a utils module

runner/src/main.rs
```diff
@@ -63,11 +67,20 @@ pub enum CommandResult {
...
 
+// TODO : Enable metrics/tracing for load generator command
+// Metrics and tracing have been disabled for load generator due to memory issues.
+// Memory grows in the runner when this is enabled not making it live long enough to finish the load generation
 #[tokio::main]
 async fn main() -> Result<()> {
     let args = Cli::parse();
-    telemetry::init_tracing(Some(args.otlp_endpoint.clone())).await?;
-    let metrics_controller = telemetry::init_metrics_otlp(args.otlp_endpoint.clone()).await?;
+    if !matches!(args.command, Command::GenerateLoad(_)) {
```
dav1do, line 76: might be nice to do metrics/tracing in one statement... and using match might be simpler but it's pretty much same same but different

```rust
let metrics_controller =  match &args.command {
        Command::GenerateLoad(_) => None,
    _ => {
        telemetry::init_tracing(Some(args.otlp_endpoint.clone())).await?;
        Some(telemetry::init_metrics_otlp(args.otlp_endpoint.clone()).await?)
    }};
```

runner/src/main.rs
```diff
@@ -79,17 +92,21 @@ async fn main() -> Result<()> {
...
+        Command::GenerateLoad(opts) => simulate_load(opts).await?,
         Command::Noop => CommandResult::Success,
     };
-
-    // Flush traces and metrics before shutdown
-    shutdown_tracer_provider();
-    metrics_controller.force_flush()?;
-    drop(metrics_controller);
-    shutdown_meter_provider();
+    if !matches!(args.command, Command::GenerateLoad(_)) {
```
dav1do, line 101: rather than clone the command and do this match again, you could just check `if metrics_controller.is_some()`.. if you don't like overloading the meaning that way, you could return a tuple above with a ` let (metrics_controller, shutdown_tracing) = match .. { (None, false) } ...` or something

runner/src/main.rs
```diff
@@ -79,17 +92,21 @@ async fn main() -> Result<()> {
...
+        // Flush traces and metrics before shutdown
+        shutdown_tracer_provider();
+        if let Some(metrics_controller) = metrics_controller.clone() {
+            metrics_controller.force_flush()?;
+        }
+        drop(metrics_controller);
+        shutdown_meter_provider();
```
dav1do, line 108: the drop/shutdown should maybe be in the `if let` statement since it's not used otherwise but it doesn't look like it's caused an issue

dav1do (commented):
Nice! A lot of small comments/rust things but overall looks good.. I'm still really sad about tracing and would like to figure that out 😭

runner/src/load_generator/utils/ceramic_models_utils.rs
dav1do: actually, I changed my mind here. I think we unwrap calls that _should never fail_ and if they do, there's no point in running the test and wasting resources so we should just crash everything and get it over with

runner/src/load_generator/gen.rs
```diff
@@ -0,0 +1,233 @@
...
+ * @return The result of the simulation
+ */
+pub async fn create_model_instances_continuously(
+    stable_load_user: StableLoadUser,
+    model: StreamId,
+    duration_in_hours: u64,
+) -> Result<()> {
+    let start_time = Instant::now();
+
+    let duration = Duration::from_secs(duration_in_hours * 60 * 60);
```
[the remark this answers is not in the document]
samika98: https://github.com/rust-lang/rust/issues/120301. This is not yet a stable feature

runner/src/main.rs
dav1do: I think we should add back the `drop(metrics_controller);`. Probably doesn't matter now but it might cause some confusion when metrics are fixed and prevent them from showing up as expected.

dav1do (approved):
LGTM other than the one comment about dropping the metrics_controller. Probably need to run a cargo update to fix CI. It looks like the new clippy/rust release doesn't like the old version of time and `cargo update -p time` should do it, but updating all dependencies is probably fine... normally I'd suggest we should do that in another PR and merge it right away, but as this is close so I think doing it here is fine.

Makefile
```diff
@@ -4,7 +4,7 @@ CARGO = CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse RUSTFLAGS="--cfg tokio_unstab
 all: build check-fmt check-clippy test
 
 .PHONY: test
-test:
+test: update
```
dav1do, line 7: I don't think we want to run update as part of these tasks. Adding an update step is fine, but we should run `make update` and commit the lockfile in a commit. I suspect this will cause unstaged changes in CI that could fail the pipeline.

runner/src/load_generator/gen.rs
```diff
@@ -0,0 +1,251 @@
...
+use ceramic_core::StreamId;
+use clap::Args;
+use keramik_common::peer_info::Peer;
+use tokio::time::{Duration, Instant};
+
+// TODO : Use this to envoke a particular scenario, currently we only have one
+// so this is unused
+#[allow(dead_code)]
+#[derive(Clone, Debug)]
+pub enum WeekLongSimulationScenarios {
```
smrz2001, line 19: What about a different term instead of `WeekLong` since the run time is configurable? Maybe `LoadGenScenarios`? If so, can replace that throughout this file.

smrz2001 (approved):
Just minor comments. LGTM! 

(review the Rust-y bits as well as I could 😊)

PR #204, opened 2024-07-17
feat: lgen operator

samika98:
Features in this pr : 
1) Generate crds with vars specific to load generation
2) Add reconciliation controller for load generation specific resources
3) Change operator yaml to give custom api access for generation load generation specific crds

Tests : 

Test 1 : Create a load generation and apply it. 

1.  Create a network (similar to basic.yaml) in the docs - network.yaml
2.  Create load generation yaml for configuring load generation 

```
  # load-generator.yaml
kind: LoadGenerator
metadata:
  name: load-gen
  namespace: keramik-small
spec:
  scenario: "model-load-sync"
  runTime: 1
  image: "samika98/runner:final-1"
  imagePullPolicy: "IfNotPresent"
  throttleRequests: 100
  tasks: 10

```

Result : Success - Was able to generate load successfully against a ceramic node. Model got indexed on the ceramic peer.

### DIFF
--- a/k8s/operator/manifests/operator.yaml
+++ b/k8s/operator/manifests/operator.yaml
@@ -36,7 +36,7 @@ rules:
     resources: ["clusterroles", "clusterrolebindings"]
     verbs: ["create", "get", "patch"]
   - apiGroups: ["keramik.3box.io"]
-    resources: ["networks", "networks/status", "simulations", "simulations/status"]
+    resources: ["networks", "networks/status", "simulations", "simulations/status", "loadgenerators", "loadgenerators/status"]
     verbs: ["get", "list", "watch", "patch", "delete"]
   - apiGroups: ["monitoring.coreos.com"]
     resources: ["podmonitors"]
@@ -87,7 +87,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: keramik-operator
-  namespace: default
+  namespace: keramik
   labels:
     app: keramik-operator
     app.kubernetes.io/name: keramik-operator
--- a/k8s/operator/manifests/operator.yaml
+++ b/k8s/operator/manifests/operator.yaml
@@ -87,7 +87,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: keramik-operator
-  namespace: keramik
+  namespace: default
   labels:
     app: keramik-operator
     app.kubernetes.io/name: keramik-operator
--- a/operator/src/lgen/controller.rs
+++ b/operator/src/lgen/controller.rs
@@ -25,7 +25,7 @@ use crate::{
     labels::MANAGED_BY_LABEL_SELECTOR,
     lgen::{
         job::{job_spec, JobConfig, JobImageConfig},
-        spec::{LoadGenerator, LoadGeneratorStatus},
+        spec::{LoadGenerator, LoadGeneratorState},
     },
     simulation::controller::monitoring_ready,
     utils::Clock,
@@ -137,7 +137,7 @@ async fn reconcile_(
         status.clone()
     } else {
         // Generate new status with random name and nonce
-        LoadGeneratorStatus {
+        LoadGeneratorState {
             nonce: thread_rng().gen(),
             name: "load-gen-"
                 .chars()
--- a/operator/src/lgen/controller.rs
+++ b/operator/src/lgen/controller.rs
@@ -1,0 +1,200 @@
+use std::{sync::Arc, time::Duration};
+
+use futures::stream::StreamExt;
+use k8s_openapi::api::batch::v1::Job;
+use kube::{
+    api::{Patch, PatchParams},
+    client::Client,
+    core::object::HasSpec,
+    runtime::Controller,
+    Api,
+};
+use kube::{
+    runtime::{
+        controller::Action,
+        watcher::{self, Config},
+    },
+    Resource, ResourceExt,
+};
+use opentelemetry::{global, KeyValue};
+use rand::{distributions::Alphanumeric, thread_rng, Rng, RngCore};
+
+use tracing::{debug, error, info};
+
+use crate::{
+    labels::MANAGED_BY_LABEL_SELECTOR,
+    lgen::{
+        job::{job_spec, JobConfig, JobImageConfig},
+        spec::{LoadGenerator, LoadGeneratorStatus},
+    },
+    simulation::controller::monitoring_ready,
+    utils::Clock,
+};
+
+use crate::network::ipfs_rpc::{HttpRpcClient, IpfsRpcClient};
+
+use crate::utils::{apply_job, Context};
+
+/// The name of the load generator job.
+pub const LOAD_GENERATOR_JOB_NAME: &str = "load-gen-job";
+
+/// Handle errors during reconciliation.
+fn on_error(
+    _network: Arc<LoadGenerator>,
+    _error: &Error,
+    _context: Arc<Context<impl IpfsRpcClient, impl RngCore, impl Clock>>,
+) -> Action {
+    Action::requeue(Duration::from_secs(5))
+}
+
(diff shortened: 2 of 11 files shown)


operator/src/lgen/spec.rs
```diff
@@ -0,0 +1,39 @@
...
+    pub image_pull_policy: Option<String>,
+    /// Throttle requests (per second) for a load generator. Currently on a per-worker basis.
+    pub throttle_requests: Option<usize>,
+    /// Number of tasks to run in parallel
+    pub tasks: usize,
+}
+
+/// Status of the load generator.
+#[derive(Serialize, Deserialize, Debug, Default, PartialEq, Clone, JsonSchema)]
+pub struct LoadGeneratorStatus {
```
dav1do, line 34: nit: status seems like an odd suffix here

runner/src/main.rs
```diff
@@ -50,7 +50,7 @@ impl Command {
             Command::Simulate(_) => "simulate",
             Command::Noop => "noop",
             // TODO : After making operator changes this command will be used to generate load
```
dav1do, line 52: looks like this comment can be deleted

operator/src/lgen/controller.rs
```diff
@@ -0,0 +1,200 @@
...
+    },
+    #[error("Kube error: {source}")]
+    Kube {
+        #[from]
+        source: kube::Error,
+    },
+}
+
+/// Start a controller for the LoadGenerator CRD.
+pub async fn run() {
```
dav1do, line 66: what do you see as the future of the simulation crd? will that likely go away? seems like there is a lot of duplication here, which is fine for now but we may want to think about how to share some things if we're going to keep both around (and/or add other variations)

operator/src/lgen/job.rs
```diff
@@ -0,0 +1,183 @@
...
+            spec: Some(PodSpec {
+                hostname: Some("job".to_owned()),
+                subdomain: Some("load-gen-job".to_owned()),
+                containers: vec![Container {
+                    name: "job".to_owned(),
+                    image: Some(config.job_image_config.image),
+                    image_pull_policy: Some(config.job_image_config.image_pull_policy),
+                    command: Some(vec![
+                        "/usr/bin/keramik-runner".to_owned(),
+                        "generate-load".to_owned(),
```
dav1do, line 158: maybe I missed it, but how is this image built?

operator/src/lgen/job.rs
```diff
@@ -0,0 +1,183 @@
...
+        })
+    }
+
+    JobSpec {
+        backoff_limit: Some(1),
+        template: PodTemplateSpec {
+            metadata: Some(ObjectMeta {
+                labels: Some(BTreeMap::from_iter(vec![(
+                    "name".to_owned(),
+                    "load-gen-job".to_owned(),
```
dav1do, line 145: should we use the `LOAD_GENERATOR_JOB_NAME` const for name in this file as well?

operator/src/lgen/job.rs
```diff
@@ -0,0 +1,183 @@
...
+                )])),
+                ..Default::default()
+            }),
+            spec: Some(PodSpec {
+                hostname: Some("job".to_owned()),
```
dav1do, line 150: I'm not super familiar with the hostname/subdomain settings here.. do we want something more specific?

dav1do (commented):
Mostly questions, am happy to approve if needed but was leaving the review to the folks more familiar with the k8s requirements.

operator/src/lgen/job.rs
```diff
@@ -0,0 +1,183 @@
...
+pub fn job_spec(config: JobConfig) -> JobSpec {
+    let mut env_vars = vec![
+        EnvVar {
+            name: "RUNNER_OTLP_ENDPOINT".to_owned(),
+            value: Some("http://otel:4317".to_owned()),
+            ..Default::default()
+        },
+        EnvVar {
+            name: "RUST_LOG".to_owned(),
+            value: Some("info,keramik_runner=trace".to_owned()),
```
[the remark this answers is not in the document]
samika98: This is just for future iterations. Cause we need tracing one less thing to do I guess when we enable it. We just need to figure out what's causing high memory consumption in tracing. 
For now this wont cause a memory consumption increase

operator/src/lgen/controller.rs
samika98: I see load generation completely separate from simulations. 
Simulations will be short lived, more flexible, produce burst traffic and offer way more flexibility and coverage in terms of use cases it covers. 
Load generation is for steady non bursty load on the network. 
I believe both can be used together as well. Have a steady load over the network, on top of that load we can run simulations to test out specific use-cases.

operator/src/lgen/job.rs
samika98: Naming is hard. We can probably be more specific here, load-gen-job sound better?

k8s/operator/manifests/operator.yaml
```diff
@@ -87,7 +87,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: keramik-operator
-  namespace: default
+  namespace: keramik
```
smrz2001, line 90: I believe this was intentionally default. IIUC, the operator doesn't itself run in the `keramik` namespace, though @nathanielc might have a better idea if it matters much.

operator/src/lgen/job.rs
samika98: Kubernetes uses the hostname and subdomain fields to create a fully qualified domain name (FQDN) for the pod. The FQDN will be of the form `hostname.subdomain.namespace.svc.cluster.local`
I feel given the subdomain, this should be fine. Any thoughts?

operator/src/lgen/job.rs
samika98: it's built using the dockerfile.

dav1do: We chatted about this... but I meant specifically is there any CI yet or is it all manual, and the answer is manual for now.

operator/src/lgen/controller.rs
```diff
@@ -0,0 +1,200 @@
...
+
+use tracing::{debug, error, info};
+
+use crate::{
+    labels::MANAGED_BY_LABEL_SELECTOR,
+    lgen::{
+        job::{job_spec, JobConfig, JobImageConfig},
+        spec::{LoadGenerator, LoadGeneratorState},
+    },
+    simulation::controller::monitoring_ready,
```
smrz2001, line 30: Should this be moved somewhere common? Maybe it's ok but feels odd to pull in simulation-related code in lgen.

operator/src/lgen/mod.rs
```diff
@@ -0,0 +1,9 @@
+/// Lgen controller arm module for reconciling load generator resources.
+#[cfg(feature = "controller")]
+pub mod controller;
+/// Job module for creating load generator jobs.
+#[cfg(feature = "controller")]
+pub mod job;
+/// Spec module for creating load generator specs.
+#[cfg(feature = "controller")]
+pub mod spec;
```
smrz2001, line 9: Want to also add these couple of lines at the end? Very minor, just for consistency with the other mods:
```
#[cfg(feature = "controller")]
pub use controller::run;
```

operator/src/main.rs
```diff
@@ -38,7 +38,8 @@ async fn main() -> Result<()> {
         Command::Daemon => {
             tokio::join!(
                 keramik_operator::network::run(),
-                keramik_operator::simulation::run()
+                keramik_operator::simulation::run(),
+                keramik_operator::lgen::controller::run(),
```
smrz2001, line 42: nit: if you also export `run`, you'll be able to use the same syntax as the other modes.

smrz2001 (approved):
Only minor comments from my side. LGTM!

